PR #127, opened 2025-10-09
Add initial React + Vite frontend setup

LambertoKalisto:

PR #129, opened 2025-10-10
Docker & docker-compose for local dev

### LINKED ISSUE #119: Docker & docker-compose for local dev
**Title:** infra: add Dockerfiles and docker-compose for backend, frontend, Postgres, Redis

**Description:**
Provide containerized local environment for consistent development across machines.

**Checklist**

* [x] Create `Dockerfile.backend` for Django Python app.
* [ ] Create `Dockerfile.frontend` for React app (dev mode recommended).
* [x] Add `docker-compose.yml` with services: `backend`, `frontend`, `db` (Postgres), `redis` (optional for channels/notifications).
* [x] Add environment variable examples `env.example` for backend (DB_HOST, DB_USER...).
* [x] Document `docker-compose up --build` steps in README.
* [x] Ensure migrations run in compose (`command` or docs).

**Acceptance criteria**

* `docker-compose up --build` starts containers (backend may print migration errors but should start).
* Postgres container is accessible from backend container with provided env vars.

---
**Original Issue:** https://github.com/Project-Stage-Academy/Forum-PrSt-FullStack/issues/3
**Created by:** mehalyna
**Original creation date:** 2025-09-30T17:35:34Z

butuz0:
The creation of `Dockerfile.frontend` and the frontend service has been moved to a separate task, as it depends on PR #127.
The addition of the `redis` service has been postponed, since it will not be required until the implementation of notifications/messaging.


Tonnnya (approved):
:thumbsup:

kesshich (approved):

backend/apps/common/management/commands/wait_for_db.py
```diff
@@ -0,0 +1,18 @@
...
+
+class Command(BaseCommand):
+    """Django command to wait for database."""
+
+    def handle(self, *args, **options):
+        db_up = False
+        while not db_up:
+            try:
+                self.check(databases=['default'])
+                db_up = True
```
VovaD199, line 26: Now the while not db_up loop runs without sleep, meaning it instantly retries over and over again.
This can create CPU overhead or flood logs.

_Add time.sleep(1) between retries_

backend/apps/common/views.py
```diff
@@ -5,5 +5,6 @@
 
 class GetHealth(APIView):
     """ Simple health-check endpoint to check health. """
+
```
VovaD199, line 10: It is better to write the docstring in PEP 257 format (no space before quotes, capitalized, terminated with a period)

backend/requirements.txt
VovaD199, line 1: Please check your encoding settings, as this is currently showing as a BIN-file

VovaD199 (changes requested):

butuz0:
made all the requested changes

backend/requirements.txt
butuz0: changed encoding from UTF-16LE to UTF-8

backend/apps/common/management/commands/wait_for_db.py
```diff
@@ -0,0 +1,20 @@
...
```
mehalyna: No connection attempt timeout can cause infinite loop. Add a maximum retry count or timeout.
```
max_retries = 30
retries = 0
while not db_up and retries < max_retries:
    # existing code
    retries += 1
```

backend/apps/common/management/commands/wait_for_db.py
```diff
@@ -0,0 +1,20 @@
...
+            except (Psycopg2OpError, OperationalError):
+                self.stdout.write('Waiting for database to become available...')
+                time.sleep(1)
```
mehalyna: Catching broad OperationalError may mask unrelated errors. Limit exception handling to database connection errors only.
```
except (Psycopg2OpError, OperationalError) as e:
    if 'database' in str(e):
        self.stdout.write('Waiting for database to become available...')
        time.sleep(1)
    else:
        raise
```

mehalyna: Sleep interval is hardcoded and may slow startup. Make the sleep duration configurable via an argument or environment variable.
```
sleep_time = int(os.environ.get("DB_WAIT_SLEEP", 1))
time.sleep(sleep_time)
```

backend/scripts/run.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+set -o errexit
+set -o pipefail
+set -o nounset
+
+python manage.py wait_for_db
+python manage.py migrate --no-input
```
mehalyna: Missing error handling for 'wait_for_db' and 'migrate' commands. Add exit-on-error to prevent server start if migrations fail.
```
python manage.py wait_for_db || exit 1
python manage.py migrate --no-input || exit 1
```

backend/backend/settings.py
```diff
@@ -90,18 +89,20 @@
...
 DATABASES = {
     'default': {
-        'ENGINE': 'django.db.backends.sqlite3',
-        'NAME': BASE_DIR / 'db.sqlite3',
+        'ENGINE': 'django.db.backends.postgresql',
+        'NAME': os.environ.get('DB_NAME'),
+        'USER': os.environ.get('DB_USER'),
+        'PASSWORD': os.environ.get('DB_PASSWORD'),
+        'HOST': os.environ.get('DB_HOST'),
+        'PORT': os.environ.get('DB_PORT'),
```
mehalyna: Add a default port to avoid issues if env var is missing.
```
'PORT': os.environ.get('DB_PORT', '5432'),
```

backend/env.templates
```diff
@@ -1,8 +1,11 @@
 DJANGO_SECRET_KEY=your-secret-key
 DEBUG=True
 
-DB_ENGINE=django.db.backends.sqlite3
-DB_NAME=db.sqlite3
+DB_NAME=db
+DB_USER=user
+DB_PASSWORD=password
+DB_HOST=host
+DB_PORT=5432
```
mehalyna: Placeholder values fmay cause confusion. Add comments indicating these must be replaced.
```
DB_USER=user # Replace 'user' with your DB username
DB_PASSWORD=password # Replace 'password' with your DB password
DB_HOST=host # Replace 'host' with your DB host
```

backend/apps/common/views.py
```diff
@@ -4,6 +4,7 @@
 
 
 class GetHealth(APIView):
-    """ Simple health-check endpoint to check health. """
+    """Simple health-check endpoint to check health."""
+
     def get(self, request):
-        return Response({'status':'ok'}, status=status.HTTP_200_OK)
\ No newline at end of file
+        return Response({'status': 'ok'}, status=status.HTTP_200_OK)
```
mehalyna: Health endpoint does not check database connectivity. Verify DB connection for a more robust health check.
```
from django.db import connections
connections['default'].cursor()  # raises error if not connected
return Response({'status': 'ok'}, status=status.HTTP_200_OK)
```

mehalyna (changes requested):

backend/apps/common/views.py
butuz0: Django connects to the database on its own during middleware processing, so any request will fail if the database becomes unavailable

backend/env.templates
butuz0: done

backend/backend/settings.py
butuz0: done

backend/apps/common/management/commands/wait_for_db.py
butuz0: added customizable max retries limit

backend/apps/common/management/commands/wait_for_db.py
butuz0: done. also the connection approach has been changed to the low-level psycopg2.connect()

butuz0: done

backend/scripts/run.sh
butuz0: done

backend/apps/common/views.py
mehalyna: Thanks - good point, you’re half-right. Django *will* error on normal requests if the DB is unavailable **but** that behavior is not sufficient for robust liveness/readiness checks (and for K8s readiness probes / uptime dashboards). We still need an explicit, lightweight DB check for *health* endpoints and probe semantics.

Why explicit DB checks are needed:
- Kubernetes readiness should fail fast when DB is unreachable so traffic is not routed to a pod that can’t serve DB-backed endpoints. Relying on “Django will fail when it tries to hit the DB” is reactive and can route traffic to a pod that will immediately error for real users.
- Health dashboards / uptime monitoring expect a deterministic answer: can the app reach its DB right now? That is different from whether some later middleware will raise.
- Controlled behavior - we can return 503 with a meaningful message rather than an unhandled 500 with stack traces.
- Avoid noisy errors: the health check can be rate-limited / cached (TTL) so we don’t hammer the DB with probe traffic.

mehalyna (commented):
You're correct that normal requests will fail if DB is down, but we still need an explicit, deterministic DB check for readiness/liveness and monitoring.
You can add the readiness endpoint (or extend the PR) with the ensure_connection() + SELECT 1 pattern:  short cache TTL, unit tests for the positive and negative cases, and a brief ops doc describing probe usage.
Consequently, we can leave this issue on your own.

backend/apps/common/views.py
butuz0: You are right, i apologize. I have added an explicit db connection check

VovaD199 (approved):

mehalyna (approved):

PR #141, opened 2025-10-13
Implement User Django model + migrations

butuz0:

PR #152, opened 2025-10-14
Implements the GET /api/startups/{id}/ endpoint #64

Domanov-Vitalii:
Linked issue - https://github.com/Project-Stage-Academy/UA-4252-Python-FS/issues/64

Adds a endpoint `GET /api/startups/{id}/` to retrieve a startup's public profile data.

- Implements `StartupPublicProfileSerializer` to format the response based on the existing `StartupProfile` model.
- The serializer dynamically calculates `followers_count` and parses `tags`.
- Adds `StartupPublicProfileViewSet` to handle the `retrieve` action.
- Registers the new URL route.
- Includes comprehensive unit tests for the new endpoint, covering success cases, 404 errors, and response schema validation.


backend/apps/startups/tests.py
```diff
@@ -172,3 +178,98 @@ def test_foreign_key_relations(self):
         saved = SavedStartup.objects.create(**self.valid_data)
         self.assertEqual(saved.investor.user.email, "investor@example.com")
         self.assertEqual(saved.startup.user.email, "owner@smartvision.ai")
+
+
+class StartupPublicProfileAPITest(APITestCase):
+    def setUp(self):
+        self.user1 = User.objects.create(
+            username="startupuser1",
```
butuz0: User does not have username field. The User model no longer includes the username field. The final version of the custom user model was approved in PR #141

backend/apps/common/urls.py
```diff
@@ -1,6 +1,8 @@
 from django.urls import path
 from . import views
+from apps.startups.views import StartupPublicProfileViewSet
 
 urlpatterns = [
     path('health/', views.GetHealth.as_view()),
+    path('startups/<int:id>/', StartupPublicProfileViewSet.as_view({'get': 'retrieve'}), name='startupprofile-detail'),
```
butuz0: The common app is designed to contain shared functionality used across the entire project.
Since this endpoint belongs to the startups domain, it should have its own urls.py inside the startups app rather than being added to common/urls.py.

backend/apps/startups/tests.py
butuz0: The User model no longer includes the username field. The final version of the custom user model was approved in PR #141

butuz0 (changes requested):

butuz0 (approved):

redrembo24680 (approved):

backend/apps/common/urls.py
```diff
@@ -1,5 +1,6 @@
 from django.urls import path
 from . import views
+from apps.startups.views import StartupPublicProfileViewSet
```
VovaD199, line 3: This import is not used, so it should be removed.
Otherwise, the linter (flake8, pylint) or CI may indicate an “unused import”.

backend/apps/startups/serializers.py
```diff
@@ -0,0 +1,39 @@
...
+            return request.build_absolute_uri(obj.logo.url)
+        return None
+
+    def get_tags(self, obj):
+        if obj.partners_brands:
+            return [tag.strip() for tag in obj.partners_brands.split(',') if tag.strip()]
+        return []
+
+    def get_followers_count(self, obj):
+        return obj.savedstartup_set.count()
```
VovaD199: Method get_followers_count

    def get_followers_count(self, obj):
        return obj.savedstartup_set.count()
Works, but:
.count() executes an SQL query. If this serializer processes many objects in the list, it creates N+1 queries.
This is not optimal for a large number of startups.
Solution:
You can make a preliminary annotation in the viewset:

    queryset = StartupProfile.objects.annotate(followers_count=Count('savedstartup'))

and in the serializer simply:

    followers_count = serializers.IntegerField(read_only=True)
— this will significantly reduce the number of SQL queries.

backend/apps/startups/urls.py
```diff
@@ -0,0 +1,7 @@
+from django.urls import path
+from . import views
```
VovaD199: Excess imports

backend/apps/startups/urls.py
```diff
@@ -0,0 +1,7 @@
...
+from apps.startups.views import StartupPublicProfileViewSet
+
```
VovaD199, line 3: Suboptimal approach to ViewSet

If you are already using ViewSet, it is usually registered via DRF Router, rather than manually via path.
This simplifies the code, provides automatic routes (list, retrieve, create, update, delete) and allows the API to scale more easily.

    from rest_framework.routers import DefaultRouter
    from .views import StartupPublicProfileViewSet
    
    router = DefaultRouter()
    router.register(r'startups', StartupPublicProfileViewSet, basename='startup')
    
    urlpatterns = router.urls
    
automatically generates /startups/ and /startups/<id>/;
better conforms to the REST standard;
easier to maintain when extending the API.

backend/apps/startups/urls.py
```diff
@@ -0,0 +1,7 @@
+from django.urls import path
+from . import views
+from apps.startups.views import StartupPublicProfileViewSet
```
VovaD199: Now the import is done directly:

    from apps.startups.views import StartupPublicProfileViewSet

but inside the same application (apps/startups).
You can do it simpler:

    from .views import StartupPublicProfileViewSet

— this will reduce dependencies and make the code more portable.

backend/apps/startups/views.py
```diff
@@ -1,3 +1,8 @@
-from django.shortcuts import render
+from rest_framework import viewsets, mixins
+from .models import StartupProfile
+from .serializers import StartupPublicProfileSerializer
 
-# Create your views here.
+class StartupPublicProfileViewSet(mixins.RetrieveModelMixin, viewsets.GenericViewSet):
+    queryset = StartupProfile.objects.all()
+    serializer_class = StartupPublicProfileSerializer
+    lookup_field = 'id'
```
VovaD199: lookup_field = 'id'
DRF uses the pk field by default.
If the StartupProfile model has a regular id field = models.AutoField(primary_key=True) - there is no point in changing it.
It is recommended to simply remove this line to leave the default behavior:

VovaD199 (changes requested):
For clarity and auto-generation of Swagger/DRF Docs, please add short docstrings.

VovaD199 (approved):

PR #154, opened 2025-10-15
backend: implement POST /api/auth/register/ for startup and investor

Tonnnya:
backend: implement POST /api/auth/register/ for startup and investor accounts

- Add registration endpoint with role-specific fields (startup/investor)
- Create inactive users with email verification
- Generate and send verification emails with tokens
- Implement anti-enumeration policy (always return 201)
- Add server-side password validation
- Create StartupProfile/InvestorProfile based on role
- Add email verification endpoint
- Add comprehensive unit tests (11 tests covering happy path, validation, duplicates)


backend/apps/authentication/models.py
```diff
@@ -0,0 +1,79 @@
...
+    created_at = models.DateTimeField(auto_now_add=True)
+    updated_at = models.DateTimeField(auto_now=True)
+
+    class Meta:
+        verbose_name = "Startup Profile"
+        verbose_name_plural = "Startup Profiles"
+
+    def __str__(self):
+        return f"{self.company_name} ({self.user.email})"
+
```
VovaD199: If apps.startups has its own StartupProfile model, this may cause a naming conflict - check to make sure there are no duplicates.
(Your previous files from apps/startups/models.py and apps/authentication/models.py contain two different StartupProfiles - you need to merge or rename one of them.)

backend/apps/authentication/serializers.py
```diff
@@ -0,0 +1,64 @@
+from rest_framework import serializers
+from django.contrib.auth.password_validation import validate_password
+from .models import User, StartupProfile, InvestorProfile
+
+
+class RegistrationSerializer(serializers.Serializer):
+    email = serializers.EmailField(required=True)
+    password = serializers.CharField(required=True, write_only=True, validators=[validate_password])
+    role = serializers.ChoiceField(choices=['startup', 'investor'], required=True)
```
VovaD199: ChoiceField(choices=['startup', 'investor']) is not a valid form — choices must be a list of tuples.

    ROLE_CHOICES = [
        ('startup', 'Startup'),
        ('investor', 'Investor'),
    ]
    
    role = serializers.ChoiceField(choices=ROLE_CHOICES, required=True)

backend/apps/authentication/serializers.py
```diff
@@ -0,0 +1,64 @@
...
+                raise serializers.ValidationError({
+                    'company_name': 'This field is required for startups'
+                })
+
+
+        if role == 'investor':
+            if not data.get('investment_range_min'):
+                raise serializers.ValidationError({'investment_range_min': 'This field is required for investors'})
+        return data
+
```
VovaD199, line 54: can make the code cleaner

    if role == 'startup' and not data.get('company_name'):
        raise serializers.ValidationError({'company_name': 'Required for startups'})
    elif role == 'investor' and not data.get('investment_range_min'):
        raise serializers.ValidationError({'investment_range_min': 'Required for investors'})

backend/apps/authentication/serializers.py
```diff
@@ -0,0 +1,64 @@
...
+        password = validated_data['password']
+        role = validated_data['role']
+
+        user = User.objects.create_user(
+            email=email,
+            password=password,
+            is_active=False,
+            username=email
+        )
+
```
VovaD199, line 80: It is worth adding validation to the duplicate email:

    if User.objects.filter(email=email).exists():
        raise serializers.ValidationError({'email': 'This email is already registered.'})

backend/apps/authentication/views.py
```diff
@@ -0,0 +1,46 @@
...
+
+        user = serializer.save()
+
+        token = default_token_generator.make_token(user)
+        uid = urlsafe_base64_encode(force_bytes(user.pk))
+
+        verification_link = f'https://mail.google.com/api/auth/verify/{uid}/{token}/'
+        send_mail(
+            subject='Verify your email',
+            message=f'Please, verify your email by clicking: {verification_link}'),
```
VovaD199: message=f'Please, verify your email by clicking: {verification_link}'),

There are incorrect parentheses and commas here.

VovaD199 (changes requested):
For clarity and auto-generation of Swagger/DRF Docs, please add short docstrings.

backend/apps/authentication/views.py
```diff
@@ -0,0 +1,46 @@
...
+            return Response({
+                'detail': 'Verification email sent.'
+            }, status=status.HTTP_200_OK)
+
+        user = serializer.save()
+
+        token = default_token_generator.make_token(user)
+        uid = urlsafe_base64_encode(force_bytes(user.pk))
+
+        verification_link = f'https://mail.google.com/api/auth/verify/{uid}/{token}/'
```
butuz0: I believe that the verification link should point to frontend or backend domain, where the user will actually confirm their email.
verification_link = f'{settings.FRONTEND_URL}/verify-email/{uid}/{token}/'

backend/apps/authentication/models.py
```diff
@@ -0,0 +1,79 @@
...
+        verbose_name = "User"
+        verbose_name_plural = "Users"
+
+    def __str__(self):
+        return self.email
+
+    def save(self, *args, **kwargs):
+        if not self.username:
+            self.username = self.email
+        super().save(*args, **kwargs)
```
butuz0: User model has already been created in PR #141. You can refer to it using the following method:
from django contrib auth import get_user_model

backend/backend/settings.py
```diff
@@ -143,4 +142,14 @@
...
+
+EMAIL_BACKEND = os.environ.get('EMAIL_BACKEND', 'django.core.mail.backends.console.EmailBackend')
+EMAIL_HOST = os.environ.get('EMAIL_HOST', 'smtp.sendgrid.net')
+EMAIL_PORT = int(os.environ.get('EMAIL_PORT', 587))
+EMAIL_USE_TLS = os.environ.get('EMAIL_USE_TLS', 'True').lower() == 'true'
+EMAIL_HOST_USER = os.environ.get('EMAIL_HOST_USER', '')
+EMAIL_HOST_PASSWORD = os.environ.get('EMAIL_HOST_PASSWORD', '')
+DEFAULT_FROM_EMAIL = os.environ.get('DEFAULT_FROM_EMAIL', 'noreply@siskidomain.com')
+
+AUTH_USER_MODEL = 'authentication.User'
```
butuz0: AUTH_USER_MODEL has been set to 'users.User' in PR #141.

backend/apps/authentication/models.py
```diff
@@ -0,0 +1,79 @@
+from django.contrib.auth.models import AbstractUser
+from django.db import models
+
+
+class User(AbstractUser):
+    email = models.EmailField(unique=True, verbose_name="Email address")
+    is_verified = models.BooleanField(default=False, verbose_name="Email verified")
```
butuz0: Instead of adding a separate is_verified field, we could use existing is_active flag for email verification purposes to keep the model simpler.
It already represents whether a user account is active or not, so you can keep new users inactive until they verify their email.

backend/apps/authentication/serializers.py
butuz0, line 1: It might be cleaner to have separate, independent serializers like UserSerializer for creating the user itself and separate serializers for StartupProfile and InvestorProfile. 
This way each one focuses only on its own model and fields, and can be reused beyond registration. 
You could create the user using UserSerializer, and depending on the role invoke the appropriate profile serializer.

butuz0 (changes requested):

backend/apps/authentication/models.py
Tonnnya: Done

backend/apps/authentication/serializers.py
Tonnnya: Done

backend/apps/authentication/serializers.py
Tonnnya: Done

backend/apps/authentication/serializers.py
Tonnnya: Done

backend/apps/authentication/views.py
Tonnnya: Fixed

backend/apps/authentication/views.py
Tonnnya: Done

backend/apps/authentication/views.py
```diff
@@ -0,0 +1,99 @@
...
+        Returns:
+            201: User created, verification email sent
+            400: Validation errors
+        """
+
+        serializer = RegistrationSerializer(data=request.data)
+
+        if not serializer.is_valid():
+            return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
+
```
VovaD199, line 50: Although it is declared in the docstring, the code does not check for an existing email before calling serializer.save().
    
    email = request.data.get('email')
    existing_user = User.objects.filter(email=email).first()
    if existing_user:
        # just return 201 without creating a new user
        return Response({
            'detail': 'Verification email sent.'
        }, status=status.HTTP_201_CREATED)


This will align the logic with the test_duplicate_email_anti_enumeration test.

backend/apps/authentication/views.py
```diff
@@ -0,0 +1,99 @@
...
+        Returns:
+            200: Email verified successfully
+            400: Invalid or expired token
+        """
+        try:
+            user_id = urlsafe_base64_decode(uid).decode()
+            user = User.objects.get(pk=user_id)
+
+            if default_token_generator.check_token(user, token):
+                user.is_active = True
```
VovaD199, line 105: It is necessary to add:

    user.is_verified = True

backend/apps/authentication/views.py
```diff
@@ -0,0 +1,99 @@
...
+from django.utils.http import urlsafe_base64_encode, urlsafe_base64_decode
+from django.utils.encoding import force_bytes
+
+from django.contrib.auth import get_user_model
+from django.conf import settings
+
+from .serializers import RegistrationSerializer
+from apps.startups.models import StartupProfile
+from apps.investors.models import InvestorProfile
+
```
VovaD199, line 13: Check if these imports are used in this code.
These are probably leftovers from a previous version - they can be removed to avoid confusion.

    from apps.startups.models import StartupProfile
    from apps.investors.models import InvestorProfile

backend/apps/authentication/views.py
```diff
@@ -0,0 +1,99 @@
...
+        verification_link = f'{settings.FRONTEND_URL}/verify/{uid}/{token}/'
+
+        send_mail(
+            subject='Verify your email',
+            message=f'Please, verify your email by clicking: {verification_link}',
+            from_email=settings.DEFAULT_FROM_EMAIL,
+            recipient_list=[user.email],
+            fail_silently=False,
+        )
+
```
VovaD199, line 71: If SMTP is not configured or the server is temporarily unavailable, the API will throw an SMTPException.
This will break the registration process.
It is better to wrap the call:
    
    try:
        send_mail(...)
    except Exception as e:
        # логувати, але не розкривати внутрішню помилку клієнту
        logger.error(f"Email sending failed: {e}")

docker-compose.yml
```diff
@@ -10,6 +10,17 @@ services:
...
+      # Email settings for MailHog
+      - EMAIL_BACKEND=django.core.mail.backends.smtp.EmailBackend
+      - EMAIL_HOST=mailhog
+      - EMAIL_PORT=1025
+      - EMAIL_USE_TLS=False
+      - EMAIL_HOST_USER=
+      - EMAIL_HOST_PASSWORD=
+      - DEFAULT_FROM_EMAIL=noreply@siskidomain.com
+      - FRONTEND_URL=http://localhost:3000
+
```
VovaD199, line 38: EMAIL_BACKEND = os.getenv("EMAIL_BACKEND", "django.core.mail.backends.console.EmailBackend")
    EMAIL_HOST = os.getenv("EMAIL_HOST", "localhost")
    EMAIL_PORT = int(os.getenv("EMAIL_PORT", 1025))
    EMAIL_USE_TLS = os.getenv("EMAIL_USE_TLS", "False") == "True"
    EMAIL_HOST_USER = os.getenv("EMAIL_HOST_USER", "")
    EMAIL_HOST_PASSWORD = os.getenv("EMAIL_HOST_PASSWORD", "")
    DEFAULT_FROM_EMAIL = os.getenv("DEFAULT_FROM_EMAIL", "noreply@example.com")
    FRONTEND_URL = os.getenv("FRONTEND_URL", "http://localhost:3000")

VovaD199 (changes requested):

backend/apps/authentication/views.py
Tonnnya: Looking at our current User model, I only see the standard `is_active` field inherited from `AbstractUser`. There's no separate `is_verified` field at the moment. 
@butuz0 also commented on this and wrote that we don't need the 'is_verified'

backend/apps/authentication/views.py
```diff
@@ -0,0 +1,99 @@
...
+    def get(self, request, uid, token):
+        """
+        Verify user's email address.
+        Returns:
+            200: Email verified successfully
+            400: Invalid or expired token
+        """
+        try:
+            user_id = urlsafe_base64_decode(uid).decode()
+            user = User.objects.get(pk=user_id)
```
kesshich, line 97: Email verification token should be single-use (according to task #73), so consider adding check if user is already verified.

```
    if user.is_active:
        return Response({'detail': 'Account already verified.'}, status=status.HTTP_400_BAD_REQUEST)
```

docker-compose.yml
Tonnnya: Currently, I'm using `os.environ.get()` in settings.py,  which is functionally equivalent to `os.getenv()`.
Should I change it?

docker-compose.yml
```diff
@@ -10,6 +10,17 @@ services:
...
     env_file:
       - ./backend/.env
+    environment:
+      # Email settings for MailHog
+      - EMAIL_BACKEND=django.core.mail.backends.smtp.EmailBackend
+      - EMAIL_HOST=mailhog
+      - EMAIL_PORT=1025
+      - EMAIL_USE_TLS=False
+      - EMAIL_HOST_USER=
+      - EMAIL_HOST_PASSWORD=
```
VovaD199, line 35: But I have a question, where will the User and Password values ​​come from?

VovaD199 (approved):

butuz0 (approved):

kesshich (approved):

docker-compose.yml
Tonnnya: @VovaD199 
For MailHog (local development), these fields should remain empty - no credentials required! 
For production (Gmail, SendGrid), we need to add real credentials

PR #156, opened 2025-10-16
Task 63

redrembo24680:


LambertoKalisto (approved):

kesshich (approved):

backend/apps/projects/urls.py
```diff
@@ -0,0 +1,10 @@
+from django.urls import path, include
+from rest_framework.routers import DefaultRouter
+from .views import ProjectViewSet
+
+router = DefaultRouter()
+router.register(r'startups/(?P<startup_id>\d+)/projects', ProjectViewSet, basename='projects')
```
mehalyna: You call router.register(r'startups/(?P<startup_id>\d+)/projects', ProjectViewSet, basename='projects'). DRF's DefaultRouter.register expects a simple prefix (no embedded slashes/regex). Using a regex-like prefix will not produce the intended nested routes and will break URL resolution and reverse lookups, likely resulting in 500s or missing routes.

Use a simple prefix and either:
- create a top-level route and filter by query param, or
- use django-rest-framework-nested-routers to create nested routing, or
- register projects and create a custom action or a path-based include. Example using nested routers (recommended):
```
from rest_framework_nested import routers
router = routers.SimpleRouter()
router.register(r'projects', ProjectViewSet, basename='projects')
projects_router = routers.NestedSimpleRouter(router, r'startups', lookup='startup')
projects_router.register(r'projects', ProjectViewSet, basename='startup-projects')
urlpatterns = router.urls + projects_router.urls
```

backend/apps/projects/models.py
```diff
@@ -11,17 +12,24 @@ class Project(models.Model):
...
     tags = models.TextField()
     visibility = models.CharField(max_length=20, default="public")
     created_at = models.DateTimeField(auto_now_add=True)
     updated_at = models.DateTimeField(auto_now=True)
 
+    def clean(self):
+        if self.target_amount < 0:
+            raise ValidationError('Target amount cannot be negative.')
+        if self.raised_amount < 0:
+            raise ValidationError('Raised amount cannot be negative.')
```
mehalyna, line 31: You implemented a clean() that raises ValidationError for negative amounts but didn't ensure it's called on save. Using Project.objects.create(...) bypasses clean() so invalid data can be persisted. Tests in this PR even create instances via objects.create.

Enforce validation by overriding save() to call full_clean() before saving, or add field-level validators on the DecimalFields. Example:
```
def save(self, *args, **kwargs):
    self.full_clean()
    super().save(*args, **kwargs)
```

backend/backend/urls.py
```diff
@@ -22,4 +22,5 @@
     path('api/startups/', include('apps.startups.urls')),
     path('api/', include('api.authorization.urls')),
     path('common/', include('apps.common.urls')),
+    path('api/', include('apps.projects.urls')),
```
mehalyna, line 29: You added path('api/', include('apps.projects.urls')). There are already API routes registered under 'api/' (e.g., 'api/startups/' above). Including the router which registers startups/.../projects under the same 'api/' prefix may lead to confusing/duplicative routes and brittle ordering-dependence.

Register the projects app URLs consistently with the rest of the API (e.g., include only apps.projects.urls at path('api/projects/', ...)) or ensure apps.projects.urls exports only non-overlapping paths. Document expected final paths and update tests accordingly.

backend/apps/projects/views.py
```diff
@@ -1,3 +1,15 @@
...
+from .serializers import ProjectSerializer
+from .pagination import ProjectPagination
+
+class ProjectViewSet(viewsets.ModelViewSet):
+    serializer_class = ProjectSerializer
+    pagination_class = ProjectPagination 
+
+    def get_queryset(self):
+        startup_id = self.kwargs['startup_id']
+        return Project.objects.filter(startup__id=startup_id).order_by('id')
```
mehalyna: get_queryset expects self.kwargs['startup_id']. Given the invalid router registration (see prior comment), startup_id may never be provided, causing a KeyError and 500 errors when resolving requests.

Fix routing (use nested routers or explicit path) so startup_id is passed, and defensively access self.kwargs.get('startup_id') and return an appropriate empty queryset or raise an Http404 if missing. *Example:*
```
startup_id = self.kwargs.get('startup_id')
if not startup_id:
    return Project.objects.none()
```

backend/apps/projects/serializers.py
mehalyna, line 1: The serializer exposes only ['id','title','status','thumbnail','short_desc']. The model requires startup, slug, description, target_amount, etc. Using ModelViewSet create/update endpoints will fail (or allow incomplete invalid objects) because required fields are missing or not writable.

Either:
- Make the serializer explicit about which fields are required and implement create()/update() to set startup from the URL and validate required fields, or
- Use a separate ReadSerializer for listing and a WriteSerializer (with required fields) for create/update. Example: include startup as read_only=True and set def perform_create(self, serializer): serializer.save(startup=...) in the viewset.

backend/apps/projects/tests/test_model.py
```diff
@@ -0,0 +1,94 @@
...
+        self.assertEqual(self.project.thumbnail, "https://example.com/image.jpg")
+        self.assertEqual(self.project.tags, "test, example")
+        self.assertEqual(self.project.visibility, "public")
+        self.assertEqual(self.project.startup, self.startup)
+        self.assertEqual(self.project.slug, "test-project")
+
+    def test_invalid_target_amount_and_raised_amount(self):
+        with self.assertRaises(ValidationError):
+            self.project.target_amount = -1000.00
+            self.project.clean()
```
mehalyna, line 65: Tests call self.project.target_amount = -1000.00; self.project.clean() to check invalid values, but project creation in setUp uses objects.create that won't call clean() so tests and app behavior differ. This reduces test realism and could hide bugs when save bypasses validation.

Make tests assert behavior that matches application behavior: either call full_clean() / save() in setUp or update model to call full_clean() in save() (see earlier suggestion), and write additional tests that attempt to persist invalid values and expect failures.

backend/apps/projects/views.py
mehalyna, line 1: The ProjectViewSet lacks any permission_classes or authentication_classes. Exposing a ModelViewSet without auth could allow unauthorized creation/modification of projects if these endpoints are enabled.

Add appropriate DRF permissions (e.g., IsAuthenticated for write, custom permissions to ensure only startup owners can modify) and document intended access control. Example:
```
from rest_framework.permissions import IsAuthenticatedOrReadOnly
permission_classes = [IsAuthenticatedOrReadOnly]
```

mehalyna (approved):

PR #160, opened 2025-10-19
Resend verification/redis throttle

kesshich:
Implemented task #73 (part of it was solved with previous task). Changes throttle logic to Redis throttle.


backend/api/authorization/views.py
```diff
@@ -66,4 +78,42 @@ def post(self, request):
...
+            if user.is_active:
+                return Response({'detail':'Account was already verified'}, status=status.HTTP_400_BAD_REQUEST)
+
+        except User.DoesNotExist:
+            return Response(status=status.HTTP_400_BAD_REQUEST) # 400 for test
+
+        token = default_token_generator.make_token(user)
+        uid = urlsafe_base64_encode(force_bytes(user.id))
+
+        verification_link = f'http://{settings.FRONTEND_URL}/api/auth/verify-email/{uid}/{token}/'
```
mehalyna: verification_link is built as f'http://{settings.FRONTEND_URL}/...' while FRONTEND_URL already includes a scheme (e.g. "http://localhost/..."), producing URLs like "[http://http://localhost](http://http//localhost)..." which will break verification links and user flows.

Do not prepend "http://" to FRONTEND_URL; ensure FRONTEND_URL contains the full base URL and use it directly.
```
verification_link = f'{settings.FRONTEND_URL}/api/auth/verify-email/{uid}/{token}/'
```
Or use urljoin for safety:
```
from urllib.parse import urljoin
verification_link = urljoin(settings.FRONTEND_URL, f'/api/auth/verify-email/{uid}/{token}/')
```

backend/api/authorization/throttling.py
```diff
@@ -0,0 +1,21 @@
+from rest_framework.throttling import BaseThrottle
+from django.core.cache import cache
+
+class CommonRedisThrottle(BaseThrottle):
+    def get_cache_key(self, request):
+        ident = self.get_ident(request)
+        return f"throttle:{ident}"
+
+    def allow_request(self, request, view):
+        rate, duration = 50, 60 # For running tests change rate.
```
mehalyna: rate, duration = 50, 60 conflicts with test_resend_throttled() which performs 6 posts and expects the 7th to 429 - the current rate will never throttle in the test, causing failing tests or hiding intended behavior.

Make rate configurable (via settings or view attribute) and align default with tests. 
```
from django.conf import settings
RATE = getattr(settings, 'COMMON_REDIS_THROTTLE_RATE', 6)
DURATION = getattr(settings, 'COMMON_REDIS_THROTTLE_DURATION', 60)
```
Use RATE/DURATION in allow_request.

backend/backend/settings.py
```diff
@@ -141,6 +141,32 @@
...
+    "default": {
+        "BACKEND": "django_redis.cache.RedisCache",
+        "LOCATION": "redis://redis:6379/1",
+        "OPTIONS": {
+            "CLIENT_CLASS": "django_redis.client.DefaultClient"
+        }
+    }
+}
+
+if 'test' in sys.argv:
```
mehalyna: Detecting test runs with 'test' in sys.argv is fragile (won't work for pytest runs, CI runners, or custom test runners). If the conditional fails, tests will use Redis cache (which may not be available) leading to flaky failures.

Use Django test settings, an environment variable, or a dedicated test settings file (e.g., DJANGO_SETTINGS_MODULE pointing to settings_test), or detect RUN_MAIN/PYTEST_CURRENT_TEST. *Example:*
```
if os.environ.get('DJANGO_TEST', '0') == '1':
    ...
```
And set DJANGO_TEST in CI/test runners.

backend/api/authorization/views.py
```diff
@@ -66,4 +78,42 @@ def post(self, request):
...
+    Generates access token for link,
+    sends mail with verification link.
+    """
+    throttle_classes = [CommonRedisThrottle]
+
+    def post(self, request):
+        email = request.data.get('email')
+
+        if not email:
+            return Response(status=status.HTTP_400_BAD_REQUEST)
```
mehalyna: The view returns 400 when the email is not found (except User.DoesNotExist: return 400), while returning 200 on success - this allows attackers to enumerate registered emails.

Return a generic success response (e.g., 200 OK with a non-disclosing message) regardless of whether the email exists. If you must distinguish for UX, ensure additional anti-enumeration controls (rate limiting + CAPTCHAs). 
```
# Always return 200 to avoid account enumeration
return Response({'detail': 'If an account with this email exists, a verification email has been sent.'}, status=200)
```

backend/api/authorization/views.py
```diff
@@ -66,4 +78,42 @@ def post(self, request):
...
+
+        verification_link = f'http://{settings.FRONTEND_URL}/api/auth/verify-email/{uid}/{token}/'
+
+        send_mail(
+            subject='Verify your email',
+            message=f'Please, verify your email by clicking: {verification_link}',
+            from_email=settings.DEFAULT_FROM_EMAIL,
+            recipient_list=[user.email],
+            fail_silently=False,
+        )
```
mehalyna: send_mail is invoked with fail_silently=False; if email backend or settings are misconfigured, this will raise and return 500 to clients. Tests may unexpectedly fail depending on environment.

Wrap send_mail in try/except and log failures; return a 200 with a generic message to callers. For tests, mock send_mail. 
```
try:
    send_mail(...)
except Exception:
    logger.exception("Failed to send verification email")
```

backend/api/authorization/throttling.py
```diff
@@ -0,0 +1,21 @@
+from rest_framework.throttling import BaseThrottle
+from django.core.cache import cache
+
+class CommonRedisThrottle(BaseThrottle):
+    def get_cache_key(self, request):
+        ident = self.get_ident(request)
+        return f"throttle:{ident}"
```
mehalyna: Key is throttle:{ident} only by ident (IP), so requests to different endpoints or APIs share the same counter causing unexpected throttling across unrelated endpoints.

Include view name, remote user, or request.path in the key. 
```
key = f"throttle:{request.path}:{ident}"
```

backend/api/authorization/throttling.py
```diff
@@ -0,0 +1,21 @@
...
+        rate, duration = 50, 60 # For running tests change rate.
+        key = self.get_cache_key(request)
+
+        current = cache.get(key)
+        if current is None:
+            cache.set(key, 1, timeout=duration)
+            return True
+        if current >= rate:
+            return False
+        cache.incr(key)
```
mehalyna: Some cache backends may not support incr or will raise if the existing value is non-integer; race conditions can cause incorrect counts.

Use atomic Redis INCR when using redis backend directly or handle the missing/incompatible case gracefully: set initial value with add/get_or_set, and on failure fallback to safe increment, e.g.:
```
try:
    cache.incr(key)
except Exception:
    current = cache.get(key) or 0
    cache.set(key, int(current) + 1, timeout=duration)
```

backend/backend/settings.py
```diff
@@ -141,6 +141,32 @@
...
+    CACHES['default'] = {
+        'BACKEND': 'django.core.cache.backends.locmem.LocMemCache',
+        'LOCATION': 'unique-for-testing',
+    }
+
+FRONTEND_URL = os.environ.get('FRONTEND_URL', 'http://localhost:8000')
```
mehalyna: settings.py now defaults FRONTEND_URL to 'http://localhost:8000/' while docker-compose sets FRONTEND_URL=http://localhost:3000/. This mismatch may cause confusing behavior locally and in integration tests.

Keep a single source of truth for FRONTEND_URL; prefer reading from environment only and align docker-compose and documentation. Example: set default to 3000 if that's the common dev port, or make docker-compose use 8000.

backend/api/authorization/views.py
kesshich: Done

backend/api/authorization/throttling.py
kesshich: Done

backend/api/authorization/views.py
kesshich: Done.

backend/backend/settings.py
kesshich: Done.

backend/api/authorization/throttling.py
kesshich: Done.

backend/api/authorization/throttling.py
kesshich: Done.

backend/api/authorization/views.py
kesshich: Done.

mehalyna (approved):

butuz0 (approved):

PR #161, opened 2025-10-19
Email uniqueness check 68

Yenot52:


frontend/src/components/EmailInput.tsx
```diff
@@ -0,0 +1,59 @@
...
+      if (!email) return setStatus(null);
+      setStatus('checking');
+      try {
+      await new Promise((r) => setTimeout(r, 800)); 
+      const exists = email.includes("test") || email.endsWith("@gmail.com"); 
+      setStatus(exists ? "exists" : "available");
+      } catch (err) {
+        console.error(err);
+        setStatus(null);
+      }
```
mehalyna: The debounced async function created with debounce(...) is never cancelled on unmount. Pending debounced calls (or their internal async timers) can call setStatus after the component unmounts, causing memory leaks and "setState on unmounted component" warnings/crashes.

Keep a ref to the debounced function and cancel it on unmount. *Example:*
```
const debouncedRef = useRef(debounce(async (email: string) => { ... }, 500));
useEffect(() => () => debouncedRef.current.cancel(), []);
// call debouncedRef.current(email) in handleChange
```
Or create the debounced function in useEffect and cancel in the cleanup.

frontend/src/pages/Register.tsx
```diff
@@ -1,7 +1,118 @@
...
+  const [available, setAvailable] = useState<boolean | null>(null);
+
+  const debouncedEmail = useDebounce(email, 500);
+
+  useEffect(() => {
+    const checkEmail = async () => {
+      if (!debouncedEmail || !debouncedEmail.includes("@")) {
+        setAvailable(null);
+        return;
+      }
```
mehalyna: The async function awaits checkEmailMock and then calls setAvailable and setChecking without cancelling or checking if the component is still mounted, which can lead to state updates after unmount.
Use an abort flag or AbortController pattern to ignore results after unmount:
```
useEffect(() => {
  let mounted = true;
  const checkEmail = async () => {
    ...
    const res = await checkEmailMock(debouncedEmail);
    if (!mounted) return;
    setAvailable(res.available);
  };
  checkEmail();
  return () => { mounted = false; };
}, [debouncedEmail]);
```

frontend/src/pages/Register.tsx
```diff
@@ -1,7 +1,118 @@
...
+
+  return (
+    <div className="flex flex-col items-center justify-center min-h-screen bg-gray-50">
+      <form
+        onSubmit={handleSubmit}
+        className="bg-white shadow-md rounded-2xl p-8 w-96"
+      >
+        <h2 className="text-2xl font-semibold mb-6 text-center">
+          Вхід на платформу
+        </h2>
```
mehalyna: The component is named Register but the title and submit button text read like a login page ("Вхід на платформу", button "Увійти"). This confuses users and may cause incorrect navigation/flows.

Decide whether this is Register or Login and make labels consistent. For register use "Реєстрація" and a button like "Зареєструватися". 
```
<h2 className="...">Реєстрація</h2>
<button type="submit">Зареєструватися</button>
```

frontend/src/components/EmailInput.tsx
```diff
@@ -0,0 +1,59 @@
...
+import debounce from 'lodash.debounce';
+
+interface EmailInputProps {
+  value: string;
+  onChange: (value: string) => void;
+}
+
+interface CheckEmailResponse {
+  exists: boolean;
+}
```
mehalyna, line 11: EmailInput declares interface CheckEmailResponse { exists: boolean } while Register.tsx checkEmailMock returns { available: boolean }. When the app moves to a shared API, these inconsistent response shapes will cause integration bugs.

Harmonize the contract (choose either exists or available) and add a small adapter or typed API client. Example unify to { available: boolean }:
```
interface CheckEmailResponse { available: boolean }
```

frontend/src/components/EmailInput.tsx
```diff
@@ -0,0 +1,59 @@
...
+const EmailInput: React.FC<EmailInputProps> = ({ value, onChange }) => {
+  const [status, setStatus] = useState<'checking' | 'available' | 'exists' | null>(null);
+
+  const checkEmail = useCallback(
+    debounce(async (email: string) => {
+      if (!email) return setStatus(null);
+      setStatus('checking');
+      try {
+      await new Promise((r) => setTimeout(r, 800)); 
+      const exists = email.includes("test") || email.endsWith("@gmail.com");
```
mehalyna: Existence is currently derived from email.includes("test") || email.endsWith("@gmail.com") - a hard-coded rule that is neither documented as a mock nor safe for production (false positives e.g., many real users use gmail).

Mark clearly as a mock and add a TODO to replace with API call.

frontend/src/components/EmailInput.tsx
```diff
@@ -0,0 +1,59 @@
...
+        value={value}
+        onChange={handleChange}
+        placeholder="Enter your email"
+      />
+      <div>
+        {status === 'checking' && <span>Checking email...</span>}
+        {status === 'available' && <span style={{ color: 'green' }}>Email is available</span>}
+        {status === 'exists' && (
+          <div style={{ color: 'red' }}>
+            If this is your account — try <a href="/login">Login</a> or <a href="/forgot-password">Reset password</a>
```
mehalyna: EmailInput links to /forgot-password while Register.tsx links to /reset-password. Inconsistent routes will break navigation and confuse routing decisions.

Align on a single canonical route (e.g., /reset-password) across components and, if possible, centralize route constants in a shared routes.ts.

frontend/src/pages/Register.tsx
anastasia-tebrii: Thank you for your feedback. 

The reason I wrote "Вхід на платформу" and "Увійти" is that this matches with Figma:

image

I understood that the design should strictly follow Figma, but I noticed that there is already a Login.tsx file (which I did not create).

Login.tsx

_type LoginProps = {
  title?: string;
};

export default function Login({ title = "🔐 Login Page" }: LoginProps) {
  return <h1>{title}</h1>;
}_

Now I’m a bit unsure how to proceed: should I keep "Вхід на платформу" and "Увійти" as in Figma — which might require adjusting or removing the existing Login.tsx — or would it be better to simply rename the component in Register.tsx?

Thank you for your time!

mehalyna: Oh, it's my fault! I'm so sorry 🙏
Let it be "Вхід на платформу" and "Увійти"

frontend/src/pages/Register.tsx
```diff
@@ -1,7 +1,125 @@
-type RegisterProps = {
-  title?: string;
+import { useState, useEffect } from "react";
+
```
mehalyna: The previous export signature that accepted RegisterProps (including title) was removed and replaced with export default function Register() (no props). This is a public API change for a page component and will break any callers or route wrappers that passed props (or tests that rely on that prop).
Restore the previous signature or create a migration: either keep RegisterProps and a default prop value, e.g. export default function Register({ title = "Register Page" }: RegisterProps), or update all imports/usages:
```
type RegisterProps = { title?: string };
export default function Register({ title = "📝 Register Page" }: RegisterProps) {
  // ... use title in <h2> or remove legacy usage only after all callers updated
}
```

frontend/src/components/EmailInput.tsx
```diff
@@ -0,0 +1,70 @@
+import React, { useState, useRef, useEffect } from "react";
+import debounce from "lodash.debounce";
```
mehalyna: EmailInput.tsx imports lodash.debounce while Register.tsx implements its own useDebounce. This inconsistency complicates maintenance and may increase bundle size due to an extra dependency for a small behavior.

Prefer a single debounce implementation across the codebase - either a small local useDebounce hook (tree-shakeable) or a shared util. If you keep lodash, import only lodash/debounce with proper bundler config. 
```
// hooks/useDebounce.ts
export function useDebounce<T>(value: T, delay = 500) { ... }
```

mehalyna (commented):
There are two separate implementations of email-checking/debounce (one as a debounced component in EmailInput.tsx, and a custom useDebounce + checkEmailMock in Register.tsx). This will lead to inconsistent behavior and duplicated maintenance.

Consolidate logic: extract a single reusable hook (e.g. useEmailAvailability(email)) or make Register.tsx consume the EmailInput component.
```
// useEmailAvailability.ts
export function useEmailAvailability(email: string) { ... }
```

kesshich (approved):

butuz0 (approved):

frontend/src/components/EmailInput.tsx
anastasia-tebrii: Fixed. Added a mounted flag to prevent state updates after unmount. Replaced the lodash.debounce function with a useDebounce hook.

frontend/src/pages/Register.tsx
anastasia-tebrii: Fixed. Added a mounted flag.

frontend/src/components/EmailInput.tsx
anastasia-tebrii: Fixed. Updated CheckEmailResponse in EmailInput.tsx to use { available: boolean }

frontend/src/components/EmailInput.tsx
anastasia-tebrii: Fixed. Added  // TODO: Replace mock with real API call

frontend/src/components/EmailInput.tsx
anastasia-tebrii: Fixed. Updated EmailInput

frontend/src/pages/Register.tsx
anastasia-tebrii: Fixed. Restored the RegisterProps

frontend/src/components/EmailInput.tsx
anastasia-tebrii: Fixed. Replaced lodash.debounce with the useDebounce hook.

mehalyna (approved):

PR #162, opened 2025-10-19
Backend: Password reset flow (request + confirm)

### LINKED ISSUE #99: Backend: Password reset flow (request + confirm)
**Title:** backend: implement password reset endpoints & email templates

**Description:**
Implement endpoints to request password reset (send email w/ token link) and to set a new password (confirm endpoint). Include safe tokens and expiry.

**API spec**

* `POST /api/auth/password-reset/` `{ "email": "user@x.com" }` → 200 (always return success to avoid account enumeration).
* `POST /api/auth/password-reset/confirm/` `{ "token": "<token>", "password": "NewP@ss1" }` → 200.

**Checklist**

* [x] Implement generating time-limited reset tokens (signed tokens or Django default `PasswordResetTokenGenerator`).
* [x] Implement email template for reset (HTML + plain) and a configurable sender.
* [x] Endpoint to accept token + new password and update password (validate complexity).
* [x] Unit tests for token generation, expiry, and reset success/failure.
* [ ] Add audit/log entry for password resets.

**Acceptance criteria**

* Request endpoint always returns 200 (no account enumeration).
* Valid token + password updates password and invalid/expired token returns 400.
* Email contains link with token (link template documented).
* Tests pass.

---
**Original Issue:** https://github.com/Project-Stage-Academy/Forum-PrSt-FullStack/issues/23
**Created by:** mehalyna
**Original creation date:** 2025-10-01T12:36:15Z

butuz0:


backend/apps/authentication/serializers.py
```diff
@@ -104,18 +106,62 @@ def create(self, validated_data):
...
         return user
+
+
+class PasswordResetRequestSerializer(serializers.Serializer):
+    email = serializers.EmailField()
+
+    def validate_email(self, value):
+        if not User.objects.filter(email=value).exists():
+            raise serializers.ValidationError('User with this email does not exist.')
+        return value
```
mehalyna, line 137: The serializer raises ValidationError if the email does not exist ("User with this email does not exist."), which contradicts the view's and tests' intent to always return 200 for password-reset requests (to avoid account enumeration). This will cause 400 responses for non-existent emails and fail the PR tests / produce an information leak.

Remove this existence check from the serializer. Let the view accept the email format but handle existence silently (e.g., try/except when fetching the user). *Example fix:*
```
def validate_email(self, value):
    # Validate format only; do not reveal existence
    return value
```
And in the view:
```
try:
    user = User.objects.get(email=serializer.validated_data['email'])
except User.DoesNotExist:
    user = None
if user:
    ...send email...
```

backend/apps/authentication/views.py
```diff
@@ -117,3 +124,68 @@ def get(self, request, uid, token):
...
+            uid = urlsafe_base64_encode(force_bytes(user.id))
+            reset_link = f'{settings.FRONTEND_URL}/reset-password?uid={uid}&token={token}'
+
+            send_password_reset_email(user, reset_link)
+
+        # always return success to avoid account enumeration
+        return Response(
+            {'detail': 'Password reset link sent.'},
+            status=status.HTTP_200_OK
+        )
```
mehalyna, line 165: The view always returns a 200 response regardless of serializer validity. While returning 200 for existing/non-existing emails is intentional to prevent enumeration, malformed requests (invalid email format/required fields) should return 400. Current behavior hides input errors and violates API contract.

Return 400 when serializer.is_valid() is False due to validation errors unrelated to existence. *Example:*
```
serializer = self.serializer_class(data=request.data)
if not serializer.is_valid():
    return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
# proceed with silent handling for missing user
```

backend/apps/authentication/serializers.py
```diff
@@ -104,18 +106,62 @@ def create(self, validated_data):
...
+    
+    @staticmethod
+    def _get_user_from_uid(uid: str) -> User:
+        try:
+            decoded_uid = urlsafe_base64_decode(uid).decode()
+            user = User.objects.get(id=decoded_uid)
+        except (ValueError, User.DoesNotExist):
+            raise serializers.ValidationError('Invalid user ID.')
+
+        return user
```
mehalyna, line 163: The method decodes the uid using urlsafe_base64_decode(uid).decode() and then does User.objects.get(id=decoded_uid). This is fragile: invalid base64 can raise binascii.Error/TypeError, and the decoded string may not match the User PK type (e.g., UUID field). The current except only catches ValueError and User.DoesNotExist.

Use Django utilities to safely get text and convert to the correct type, handle base64 decode errors, and support UUID PKs. *Example:*
```
from django.utils.encoding import force_str
import binascii
try:
    decoded = force_str(urlsafe_base64_decode(uid))
except (TypeError, ValueError, binascii.Error):
    raise serializers.ValidationError('Invalid UID.')
# if PK is UUID, cast: from uuid import UUID; try: pk = UUID(decoded)
user = User.objects.filter(pk=decoded).first()
if not user:
    raise serializers.ValidationError('Invalid user ID.')
```

backend/apps/authentication/serializers.py
```diff
@@ -104,18 +106,62 @@ def create(self, validated_data):
...
+
+        attrs['user'] = user
+        return attrs
+
+    def save(self, **kwargs) -> User:
+        user = self.validated_data['user']
+        new_password = self.validated_data['new_password']
+        user.set_password(new_password)
+        user.save()
+        return user
```
mehalyna, line 213: The serializer sets the new password without running Django's password validators (validate_password). This can allow weak/insecure passwords and bypass project-wide password policies.

Call validate_password(new_password, user=user) before setting it, and raise serializers.ValidationError on failure, e.g.
```
from django.contrib.auth.password_validation import validate_password
validate_password(new_password, user)
user.set_password(new_password)
user.save()
```

backend/apps/authentication/serializers.py
```diff
@@ -104,18 +106,62 @@ def create(self, validated_data):
...
+    def validate(self, attrs):
+        if attrs['new_password'] != attrs['re_new_password']:
+            raise serializers.ValidationError('Passwords do not match.')
+
+        user = self._get_user_from_uid(attrs['uid'])
+        if not default_token_generator.check_token(user, attrs['token']):
+            raise serializers.ValidationError('Invalid or expired token.')
+
+        attrs['user'] = user
+        return attrs
```
mehalyna, line 192: validate() calls _get_user_from_uid and default_token_generator.check_token; when these raise serializers.ValidationError the message may be generic. Also no distinction between invalid uid vs invalid token vs mismatch. This reduces debuggability for API consumers and tests.

Keep high-level generic messages for security but consider differentiating input-format errors (return 400 with "Malformed UID") vs auth failures; at minimum ensure correct 400 status and consistent messages for tests. Document behavior in serializer docstring.

backend/apps/authentication/tests.py
```diff
@@ -248,3 +254,118 @@ def test_nonexistent_user(self):
...
+        self.assertTrue(default_token_generator.check_token(self.user, token))
+
+    def test_token_invalid_after_login(self):
+        """Check token invalid after user successfully logs in"""
+        token = default_token_generator.make_token(self.user)
+
+        self.user.last_login = timezone.now()
+        self.user.save()
+        self.assertFalse(default_token_generator.check_token(self.user, token))
+
```
mehalyna, line 289: The test sets token_generator._now to a lambda to simulate future time; this relies on private implementation detail of the token generator and may break across Django versions.

Use time mocking (e.g., freezegun or monkeypatching time.time), or set PASSWORD_RESET_TIMEOUT to a short value and sleep/advance time accordingly. Alternatively, use PasswordResetTokenGenerator.check_token by creating a token and patching settings.PASSWORD_RESET_TIMEOUT to 0 for immediate expiry.

backend/apps/authentication/tests.py
mehalyna, line 1: Tests generate uid using urlsafe_base64_encode(force_bytes(self.user.id)). If the User model uses UUID PKs, decoded uid may need to be converted to UUID to query. Current code uses User.objects.get(id=decoded_uid) which can fail for UUID primary keys.

Evidence: Test uses force_bytes(self.user.id) and serializer decodes to string then User.objects.get(id=decoded_uid).

When decoding, detect if PK field is UUIDField and coerce decoded value to UUID before lookup. Use User._meta.pk to check field type or use filter(pk=decoded) which accepts strings for UUIDs only if properly formatted—prefer explicit cast.

backend/apps/authentication/views.py
mehalyna, line 165: The view builds reset_link using settings.FRONTEND_URL without validation; if FRONTEND_URL is missing, the view will raise AttributeError. Also DEFAULT_FROM_EMAIL may be missing causing email send to fail.

mehalyna (commented):
Inconsistent handling of non-existent emails: serializer rejects, view assumes existence in  files: `backend/apps/authentication/serializers.py` and `backend/apps/authentication/views.py`

Serializer currently rejects unknown emails but view logic assumes success and calls User.objects.get(email=...). If serializer is changed (per earlier suggestion) to allow unknown emails, the view will raise User.DoesNotExist.

Normalize behavior: either let serializer do only format validation and let view handle silent non-existence OR keep serializer checking and ensure view won't call get blindly. Preferred - serializer validates format only and view uses try/except around get (no-op if user missing).

backend/apps/authentication/serializers.py
butuz0: Done

backend/apps/authentication/views.py
butuz0: Done

backend/apps/authentication/serializers.py
butuz0: Done

backend/apps/authentication/serializers.py
butuz0: Done

backend/apps/authentication/tests.py
butuz0: Done

backend/apps/authentication/views.py
butuz0: Done

redrembo24680 (approved):

mehalyna (commented):
I didn't recognize changes or answers on all my comments. Please, provide answers why the fixes were not made

butuz0:
> I didn't recognize changes or answers on all my comments. Please, provide answers why the fixes were not made

I have added 3 new commits with changes made according to you comments. You might have accidentally opened initial versions of changed files.

mehalyna (approved):

Domanov-Vitalii (approved):

PR #171, opened 2025-10-20
logo file upload implementation for registration

Domanov-Vitalii:
linked issue - https://github.com/Project-Stage-Academy/UA-4252-Python-FS/issues/70

Checklist (backend)
Endpoint for multipart uploads: Yes, implemented.
The endpoint /api/auth/register/) accepts POST requests.
The RegistrationSerializer serializer contains a field logo = serializers.ImageField(...), which allows handling file uploads via multipart/form-data.
File type and size validation: Yes, implemented.
The RegistrationSerializer serializer has a method validate_logo, which checks the file size (up to 10 MB) and its type (jpeg, png, svg).
File storage (locally): Yes, implemented.
The StartupProfile and InvestorProfile models use models.ImageField with upload_to, which stores files locally in the media/ folder. This meets the "dev local storage" condition.
Return file URL: Partial.
After successful registration, the endpoint returns 201 CREATED with a message that the email has been sent, but does not return profile data with the logo URL. This is not a critical issue for registration.
Checklist (frontend)
File chooser, preview, upload progress: Partially.

File chooser: Yes, implemented in RegisterStartup.tsx via <input type="file">.
Preview (thumbnail): Yes, implemented. Both forms show a preview of the image after selection.
Upload progress: No, not implemented. The progress of uploading the file to the server is not displayed.
File upload: Yes, implemented.
Both registration forms RegisterStartup.tsx and RegisterInvestor.tsx use FormData to send data along with the file in a single multipart request.
Validation errors: Yes, implemented.

File type: There is validation of file size and type on the frontend and backend so that not everything is sent to the server, but checked locally.
Unit tests: Yes, implemented.


kesshich (approved):

backend/apps/authentication/serializers.py
```diff
@@ -59,6 +60,17 @@ def validate_email(self, value):
...
+    def validate_logo(self, value):
+        """ 
+        Validate logo file size and format
+        """
+        if value.size > 10485760:  # 10 MB limit
+         raise serializers.ValidationError("Logo file size cannot exceed 10 MB.")
+        valid_types = ['image/jpeg', 'image/png', 'image/svg+xml']
+        if hasattr(value, 'content_type') and value.content_type not in valid_types:
+            raise serializers.ValidationError("Logo must be JPEG, PNG, or SVG format.")
+        return value
```
mehalyna, line 92: The serializer field logo is declared required=False, allow_null=True but validate_logo immediately accesses value.size (and later value.content_type) without checking for None. If the field is omitted or null, validation will raise an AttributeError and crash user registration.

Guard early for None and empty values. Example:
```
def validate_logo(self, value):
    if value in (None, ''):
        return value
    MAX_SIZE = 10 * 1024 * 1024
    if getattr(value, "size", 0) > MAX_SIZE:
        raise serializers.ValidationError("Logo file size cannot exceed 10 MB.")
    ...
```

backend/backend/settings.py
```diff
@@ -62,6 +62,7 @@
 INSTALLED_APPS = DJANGO_APPS + THIRD_PARTY_APPS + LOCAL_APPS
 
 MIDDLEWARE = [
+    'corsheaders.middleware.CorsMiddleware',
```
mehalyna, line 67: The PR adds 'corsheaders.middleware.CorsMiddleware' to MIDDLEWARE but does not add 'corsheaders' to INSTALLED_APPS, nor any CORS settings. If django-cors-headers is not installed, startup will fail; and omitting INSTALLED_APPS/config may produce incomplete behavior.

Either remove this change, or add instructions and config: add 'corsheaders' to INSTALLED_APPS and configure allowed origins (e.g. CORS_ALLOWED_ORIGINS or CORS_ALLOW_ALL_ORIGINS), and ensure requirements.txt/Pipfile includes django-cors-headers.

frontend/src/pages/RegisterInvestor.tsx
```diff
@@ -74,21 +76,24 @@ type MultiField = "representing" | "entityType";
...
+    formDataObj.append("company_name", formData.companyName);
+    formDataObj.append("investment_range_min", isNaN(min) ? "0" : String(min));
+    formDataObj.append("investment_range_max", isNaN(max) ? "0" : String(max));
+    if (formData.logoFile) {
+      formDataObj.append("logo", formData.logoFile);
+    }
 
-      const response = await fetch("/api/auth/register/", {
+    try {
+      const response = await fetch("http://localhost:8000/api/auth/register/", {
```
mehalyna: The client now calls a hardcoded http://localhost:8000 backend. This will fail in production, in CI, and with non-local dev setups and requires CORS configuration. Previously the code used relative URLs.

Use a configurable base URL (env var, runtime config, or keep relative paths). 
```
const API_BASE = process.env.REACT_APP_API_BASE || '';
const response = await fetch(`${API_BASE}/api/auth/register/`, { method: 'POST', body: formDataObj });
```

frontend/src/pages/RegisterStartup.tsx
```diff
@@ -260,6 +286,18 @@ export default function RegisterStartup() {
...
+              <div className="field">
+                <label>Логотип компанії</label>
+                <input type="file" name="logoFile" accept="image/png, image/jpeg, image/svg+xml" onChange={handleChange}/>
+                {form.logoFile && <p>Вибрано: {form.logoFile.name}</p>}
+                {logoPreview && !errors.logoFile && (
+                  <div style={{marginTop: "8px"}}>
+                    <img src={logoPreview} alt="Logo preview" style={{maxWidth: "120px", maxHeight: "120px", borderRadius: "8px"}} />
+                  </div>
+                )}
+                {errors.logo && <p className="error-text">{errors.logo}</p>}
```
mehalyna, line 317: The component uses several different error keys: errors.logoFile, errors.logo, and errors.logoFile (in another check). This mismatch will prevent proper display/clearing of validation errors and can lead to stale error messages.

Pick a single key (e.g. errors.logo) and use consistently across validation and rendering. Update validation to set errors.logo and checks to reference errors.logo.

backend/apps/authentication/serializers.py
```diff
@@ -59,6 +60,17 @@ def validate_email(self, value):
...
+    
+    def validate_logo(self, value):
+        """ 
+        Validate logo file size and format
+        """
+        if value.size > 10485760:  # 10 MB limit
+         raise serializers.ValidationError("Logo file size cannot exceed 10 MB.")
+        valid_types = ['image/jpeg', 'image/png', 'image/svg+xml']
+        if hasattr(value, 'content_type') and value.content_type not in valid_types:
+            raise serializers.ValidationError("Logo must be JPEG, PNG, or SVG format.")
```
mehalyna: models.ImageField and DRF ImageField assume image processing (Pillow). Pillow typically cannot process SVGs, and Django may not accept SVG as a valid ImageField, causing server errors on save/validation despite this accept list.

Either restrict to raster types (image/jpeg, image/png) or change model/serializer to use FileField for SVGs and perform appropriate content validation. 
```
valid_types = ['image/jpeg', 'image/png']
# or if allowing SVG:
logo = models.FileField(upload_to='startup_logos/')
```

frontend/src/pages/RegisterInvestor.tsx
```diff
@@ -74,21 +76,24 @@ type MultiField = "representing" | "entityType";
...
+    const formDataObj = new FormData();
+    formDataObj.append("email", formData.email);
+    formDataObj.append("password", formData.password);
+    formDataObj.append("role", formData.role);
+    formDataObj.append("company_name", formData.companyName);
+    formDataObj.append("investment_range_min", isNaN(min) ? "0" : String(min));
+    formDataObj.append("investment_range_max", isNaN(max) ? "0" : String(max));
+    if (formData.logoFile) {
+      formDataObj.append("logo", formData.logoFile);
+    }
```
mehalyna, line 99: Numeric fields are appended as strings; while FormData only supports strings and files, the backend may expect integers. The server may still coerce, but this can create subtle bugs or validation mismatches.

Ensure backend serializer accepts string numbers (coercion) or cast on backend. Alternatively, explicitly format values consistently and document the contract. Example server-side serializer: IntegerField() will coerce string ints automatically; if not, update serializer.

backend/apps/authentication/serializers.py
butuz0, line 92: I suggest moving image validation logic (size, format, etc.) into reusable functions in a shared module.
For example:
```
# apps/common/validators.py
def validate_file_size(file, max_size_mb):
def validate_file_type(file, allowed_types):
```
Also it might be better to implement filename normalization to avoid issues with user-provided filenames. If the original filenames are not important for the users, you could use UUID for file names.

frontend/src/pages/RegisterInvestor.tsx
Domanov-Vitalii: class RegistrationSerializer(serializers.Serializer):
    ....
    # investor
    investment_range_min = serializers.DecimalField(max_digits=12, decimal_places=2, required=False, allow_null=True)
    investment_range_max = serializers.DecimalField(max_digits=12, decimal_places=2, required=False, allow_null=True)


DRF automatically converts strings into numbers for DecimalField and IntegerField if string is a valid number. 
For example: "10", "10.00", "10.45" will be converted, but "abc" - won't.

backend/apps/authentication/serializers.py
```diff
@@ -59,6 +61,16 @@ def validate_email(self, value):
...
+    
+    def validate_logo(self, value):
+        """ 
+        Validate logo file size and format
+        """
+        if value in (None, ''):
+            return value
+        validate_file_size(value, 10)
+        validate_file_type(value, ['image/jpeg', 'image/png'])
+        return value
```
mehalyna, line 92: The serializer calls validate_file_size / validate_file_type which raise django.core.exceptions.ValidationError. DRF serializer field validation should raise rest_framework.serializers.ValidationError; raising Django's ValidationError here can produce uncaught exceptions or return 500 instead of proper 400 validation responses.

Catch Django ValidationError and re-raise as serializers.ValidationError, or change validator functions to raise serializers.ValidationError. 
```
from rest_framework import serializers as drf_serializers
...
try:
    validate_file_size(value, 10)
    validate_file_type(value, ['image/jpeg','image/png'])
except ValidationError as e:
    raise drf_serializers.ValidationError(str(e))
```

backend/apps/common/validators.py
```diff
@@ -0,0 +1,10 @@
+from django.core.exceptions import ValidationError
+
+def validate_file_size(file, max_size_mb):
+    max_size = max_size_mb * 1024 * 1024
+    if getattr(file, "size", 0) > max_size:
+        raise ValidationError(f"Розмір файлу не може перевищувати {max_size_mb} MБ.")
+
+def validate_file_type(file, allowed_types):
+    if hasattr(file, 'content_type') and file.content_type not in allowed_types:
+        raise ValidationError(f"Файл повинен бути одним з: {', '.join(allowed_types)}.")
```
mehalyna: The validator trusts file.content_type (provided by the client) to determine allowed file types. This is easily spoofable and could allow non-image files to be uploaded, potentially introducing remote code/data risks if processed or served.

Validate image content server-side (e.g., open with Pillow or use Python's imghdr/filetype), check file signature/magic bytes, and fall back to extension only if needed. Example using Pillow:
```
from PIL import Image, UnidentifiedImageError
def validate_file_type(file, allowed_types):
    try:
        Image.open(file).verify()
    except UnidentifiedImageError:
        raise ValidationError("Invalid image file.")
```

backend/apps/common/utils.py
```diff
@@ -0,0 +1,6 @@
+
+def logo_upload_to(instance, filename):
+    ext = filename.split('.')[-1]
+    user_uuid = instance.user.id  
+    filename = f"{user_uuid}.{ext}"
+    return f"logos/{filename}"
```
mehalyna: The upload function uses instance.user.id as filename and original extension from filename.split('.')[-1]. This can cause filename collisions, reveal internal user IDs, and break if instance.user is not set or filename has no extension.

Use uuid4() or user UUID if available, sanitize/validate extension, and guard against missing instance.user. 
```
import uuid, os
def logo_upload_to(instance, filename):
    ext = os.path.splitext(filename)[1].lower() or '.png'
    uid = getattr(instance.user, 'pk', uuid.uuid4())
    return f"logos/{uid}-{uuid.uuid4().hex}{ext}"
```

backend/apps/common/validators.py
```diff
@@ -0,0 +1,10 @@
+from django.core.exceptions import ValidationError
+
+def validate_file_size(file, max_size_mb):
+    max_size = max_size_mb * 1024 * 1024
+    if getattr(file, "size", 0) > max_size:
+        raise ValidationError(f"Розмір файлу не може перевищувати {max_size_mb} MБ.")
```
mehalyna: The validators raise django.core.exceptions.ValidationError. They are reused inside DRF serializers (and possibly elsewhere) which expect DRF serializers' errors. This inconsistent exception type requires callers to convert exceptions every time.

Either document that these raise Django ValidationError and ensure all callers convert, or change these functions to raise rest_framework.serializers.ValidationError (or provide a small wrapper). Prefer keeping validators library-agnostic and let the serializer wrap exceptions

backend/apps/investors/models.py
```diff
@@ -52,7 +53,7 @@ class InvestorProfile(models.Model):
     city = models.CharField(max_length=100)
     address = models.CharField(max_length=200)
     postal_code = models.CharField(max_length=20)
-    logo = models.ImageField(upload_to='media/Investor_logos/')
+    logo = models.ImageField(upload_to=logo_upload_to, blank=True, null=True)
```
mehalyna, line 64: Changing upload_to from a static path to logo_upload_to will change where new files are stored and may not map to previously uploaded files. This requires migrations and a plan to move existing media or it will break existing references.

Create and include a Django migration, update docs, and run a migration plan to relocate existing files or retain fallback logic in logo_upload_to to resolve old paths.

backend/apps/startups/models.py
```diff
@@ -17,8 +18,7 @@ class StartupProfile(models.Model):
     city = models.CharField(max_length=100)
     address = models.CharField(max_length=255)
     postal_code = models.CharField(max_length=20)
-    # We need to craete media folder and set MEDIA_URL and MEDIA_ROOT in settings.py
-    logo = models.ImageField(upload_to='media/startup_logos/')
+    logo = models.ImageField(upload_to=logo_upload_to, blank=True, null=True)
```
mehalyna, line 30: Same as above - changing upload_to will affect where files are stored and served; missing migration or migration notes will lead to broken image URLs.

backend/apps/common/validators.py
```diff
@@ -0,0 +1,17 @@
...
+        raise serializers.ValidationError(f"Розмір файлу не може перевищувати {max_size_mb} MБ.")
+
+def drf_validate_file_type(file, allowed_types):
+    ext = os.path.splitext(file.name)[1].lower()
+    if ext not in allowed_types:
+        raise serializers.ValidationError(f"Дозволені розширення: {', '.join(allowed_types)}.")
+    try:
+        Image.open(file).verify()
+    except UnidentifiedImageError:
+        raise serializers.ValidationError("Invalid image file.")
```
mehalyna: Image.open(file).verify() advances/consumes the uploaded file stream; after verify the uploaded file pointer is at EOF which can cause saved images to be empty/corrupted or fail during later processing. Also only UnidentifiedImageError is caught — other IO errors (OSError, ValueError) may be raised and are not handled.

Seek back to beginning after verify and catch broader exceptions. 
```
try:
    Image.open(file).verify()
    file.seek(0)
except (UnidentifiedImageError, OSError, ValueError) as e:
    raise serializers.ValidationError("Невірний файл зображення.") from e
```
Add unit tests to ensure file can be saved after validation.

backend/apps/startups/migrations/0003_alter_startupprofile_logo.py
mehalyna, line 1: The migration import apps.common.utils imports project application code at migration runtime. This couples migrations to current project code and can break when code moves/changes (bad practice per Django docs).

Define a local helper upload_to in the migration or reference the callable by dotted string (if storage supports) or use a simple string upload path. Example: move logo_upload_to implementation into the migration file and pass that local function to AlterField.

backend/apps/startups/migrations/0004_move_logos_to_uuid.py
mehalyna, line 1: The migration uses shutil.move on MEDIA_ROOT files which can cause data loss if interrupted; it doesn't use Django storage APIs, lacks retries/atomicity, and will fail silently or partially on remote/cloud storages (S3). It also assumes files are local disk files.

Use Django's storage API (default_storage.open/save) or a safe copy-and-delete with backups, and add clear logging and fail-safe behavior. Also make this an optional management command instead of automatic migration if moving production media.
```
from django.core.files.storage import default_storage
with default_storage.open(old_name, 'rb') as f:
    content = f.read()
default_storage.save(new_name, ContentFile(content))
default_storage.delete(old_name)
```

backend/apps/startups/migrations/0006_fix_logo_paths2.py
mehalyna, line 1: There are two nearly identical migration files performing the same path-fix work. Duplicate migrations can cause confusion, duplicate work, and unexpected dependency ordering or re-run attempts.

Remove the redundant migration (keep one, ensure it has correct dependencies). Rebase to a single deterministic migration that performs the fix once.

backend/apps/common/utils.py
```diff
@@ -0,0 +1,14 @@
+import uuid
+import os
+
+
+def logo_upload_to(instance, filename):
+    ext = os.path.splitext(filename)[1].lower() or '.png'
+    user_uuid = None
+    if hasattr(instance, 'user') and hasattr(instance.user, 'id'):
```
mehalyna: The code checks hasattr(instance.user, 'id') but doesn't check whether instance.user.id is truthy. If user exists but id is None (unsaved user), filename becomes "None.ext" causing collisions and surprising behavior.

Use a safer check (if getattr(instance, 'user', None) and getattr(instance.user, 'id', None)) and fallback to uuid when id is falsy. Example:
```
user_id = getattr(getattr(instance, 'user', None), 'id', None)
user_uuid = str(user_id) if user_id else str(uuid.uuid4())
```

backend/apps/common/validators.py
mehalyna: Add unit tests that cover validation path.

mehalyna (commented):
It seems that problems are just piling up in your PR. I can explain why. It is advisable to respond to the reviewer's comments either with changes in the code, or with a reply directly under the comment. Quite often I do not notice changes in yours according to my comments. As a result, I have to review all the files every time. Please follow the order of work with comments, which was determined in the first meetings. Thanks in advance

butuz0 (approved):

backend/apps/common/validators.py
Domanov-Vitalii: def drf_validate_file_type(file, allowed_types):
    ext = os.path.splitext(file.name)[1].lower()
    if ext not in allowed_types:
        raise serializers.ValidationError(f"Дозволені розширення: {', '.join(allowed_types)}.")
    try:
        Image.open(file).verify()
        file.seek(0) # added
    except (UnidentifiedImageError, OSError, ValueError) as e: # added to provide other IO errors handled correctly
        raise serializers.ValidationError("Invalid image file.") from e # and also here
        
and also created file tests.py in backend/apps/common and place there unit-tests to verify that file can be saved after validation and other mechanisms of validators, such as: validate file size, validate file type and file content

backend/apps/common/utils.py
Domanov-Vitalii: Okay, now this func looks like: 
def logo_upload_to(instance, filename):
    ext = os.path.splitext(filename)[1].lower() or '.png'
    user_id = getattr(getattr(instance, 'user', None), 'id', None) # added
    user_uuid = str(user_id) if user_id else str(uuid.uuid4()) # edited
    filename = f"{user_uuid}{ext}"
    return os.path.join('logos', filename)
Now it works just like in the example and would never create a file with name "None.ext" that can cause collisions

backend/apps/common/validators.py
Domanov-Vitalii: Unittest was added in backend/apps/common/tests.py and covered the validation path

backend/apps/startups/migrations/0003_alter_startupprofile_logo.py
Domanov-Vitalii: So now files 0003_alter_startupprofile_logo.py and 0003_alter_investorprofile_logo.py work according to the example. 
0003_alter_investorprofile_logo.py

import os
import uuid
from django.db import migrations, models

def logo_upload_to(instance, filename):
    ... new implementation ....

class Migration(migrations.Migration):

    dependencies = [
        ('investors', '0002_initial'),
    ]

    operations = [
        migrations.AlterField(
            model_name='investorprofile',
            name='logo',
            field=models.ImageField(blank=True, null=True, upload_to=logo_upload_to), 
        ),
    ]

and the 0003_alter_startupprofile_logo.py same logic

backend/apps/startups/migrations/0006_fix_logo_paths2.py
Domanov-Vitalii: Since there was some redundancy and confusion, as well as unsafe work with files, I decided to refactor and fix the migrations for moving logos a bit.

on foto 0007, but i changed name to 0005!!!! 
A single, reliable migration (0005_consolidate_logo_fix) has been created:
It combines all the logic from previous attempts in one place, utilizing the Django Storage API's default_storage instead of shutil. The default_storage.exists(), default_storage.open(), default_storage.save(), and default_storage.delete() methods are now employed. This ensures that operations with files will work correctly regardless of where they are stored (locally or in the cloud).
Before attempting to move, the file is checked for existence. If the file does not exist, the migration logs this and continues working, rather than crashing.
Cleaned up the bad migration history.

The new migration 0005 now depends directly on 0004, which makes the story clean and logical.
image
or
image

mehalyna (approved):

mehalyna:
The failing job (ID: 53656566377) in your workflow is caused by Python code style issues, specifically related to line length and missing newline at end of file. The error messages are:

- backend/backend/settings.py:206:80: E501 line too long (84 > 79 characters)
- backend/backend/settings.py:213:80: E501 line too long (97 > 79 characters)
- backend/backend/urls.py:40:80: E501 line too long (80 > 79 characters)
- backend/backend/urls.py:40:81: W292 no newline at end of file

## Solution

1. **Shorten Lines**  
   Edit the following files and lines to ensure that no line exceeds 79 characters:
   - `backend/backend/settings.py`, lines 206 and 213
   - `backend/backend/urls.py`, line 40

   Example fix:
   ```python
   # Instead of (84 chars)
   SOME_SETTING = "this is a very long string that goes over the line length limit in flake8"
   # Split over multiple lines (using implicit string concatenation or parentheses)
   SOME_SETTING = (
       "this is a very long string that goes over the line length "
       "limit in flake8"
   )
   ```

2. **Add Newline at End of File**  
   Ensure `backend/backend/urls.py` ends with a newline (press Enter after the last line).

## Code Suggestions

For example, in `settings.py`:
```python
# Line 206 (fix long line)
DATABASES = {
    'default': {
        'ENGINE': 'django.db.backends.postgresql',
        'NAME': 'yourdbname',
        # ... other settings ...
    }
}
# If a line is too long, break it up:
'ENGINE': 'django.db.backends.postgresql',
# or use parentheses, string concatenation, or variables as needed.
```

For `urls.py`:
```python
# Line 40 (fix long line and newline issue)
urlpatterns = [
    # ... your url patterns ...
]
# Make sure there's a blank line at the end of the file.
```

## Next Steps

- Make these corrections in your PR branch.
- Commit and push the changes.
- Re-run the workflow.

These fixes will resolve the style errors causing your CI job to fail. If you need the exact code from those lines, let me know!

Issue #82, opened 2025-10-04
Restore Password (Set New Password) desktop


mehalyna:
**Short description:**
Implement the “Restore password / Set new password” desktop flow from the mockups: token parsing from link, secure token verification, new password form with client & server validation, single-use tokens, friendly UX (success, invalid/expired token, resend), logging and tests. 

---
**Original Issue:** https://github.com/Project-Stage-Academy/Forum-PrSt-FullStack/issues/40
**Created by:** mehalyna
**Original creation date:** 2025-10-01T13:02:31Z

PR #173, opened 2025-10-21
Security Docs for Restore Password

kesshich:
Initial version of .md docs for restore password #82


backend/api/authorization/tests/tests.py
```diff
@@ -123,3 +122,35 @@ def test_refresh_token_invalid(self):
...
+        resend_data = {
+            'email': self.email
+        }
+
+        for i in range(6):
+            self.client.post(self.resend_url, resend_data, format='json')
+
+        response = self.client.post(self.resend_url, resend_data, format='json')
+
+        self.assertEqual(response.status_code, status.HTTP_429_TOO_MANY_REQUESTS)
```
mehalyna, line 156: Tests in tests.py expect throttling to trigger after 6 attempts, but COMMON_REDIS_THROTTLE_RATE is set to 50 in settings. The throttle implementation reads the setting and will therefore not return 429 as tests expect.

Make test and config consistent. Either change COMMON_REDIS_THROTTLE_RATE to 6 or change the tests to override settings. For tests, prefer:
```
from django.test import override_settings
@override_settings(COMMON_REDIS_THROTTLE_RATE=6)
class ResendVerificationTests(APITestCase):
    ...
```

backend/api/authorization/views.py
```diff
@@ -66,4 +78,45 @@ def post(self, request):
...
+            if user.is_active:
+                return Response({'detail':'Account was already verified'}, status=status.HTTP_400_BAD_REQUEST)
+
+        except User.DoesNotExist:
+            return Response(status=status.HTTP_200_OK) # 400 for test
+
+        token = default_token_generator.make_token(user)
+        uid = urlsafe_base64_encode(force_bytes(user.id))
+
+        verification_link = f'{settings.FRONTEND_URL}/api/auth/verify-email/{uid}/{token}/'
```
mehalyna, line 106: Usually frontend hosts a verification route (e.g. /verify-email/<uid>/<token>), whereas /api/... implies a backend endpoint. This produces an invalid link for users or leaks an API path to the frontend.

Decide and document where verification happens. If frontend handles the click, use:
```
verification_link = f'{settings.FRONTEND_URL}/verify-email/{uid}/{token}/'
```
If backend handles it, point to backend URL (use request.build_absolute_uri or settings.BACKEND_URL).

docker-compose.yml
```diff
@@ -49,12 +48,15 @@ services:
...
     image: mailhog/mailhog:latest
     ports:
       - "1025:1025"
       - "8025:8025"
-
+  redis:
+    image: "redis:alpine"
+    hostname: redis
+    ports:
+      - "6379:6379"
```
mehalyna: ports: - "6379:6379" maps Redis to the host network. Exposing Redis is a significant security risk (unauthenticated access) for dev machines and CI runners.
Remove the host port mapping and rely on Docker network. Use:
```
redis:
  image: redis:alpine
  hostname: redis
  # remove `ports:` mapping to avoid exposing to host
```
If host access is needed in local dev, document it and secure with Redis AUTH.

docker-compose.yml
```diff
@@ -35,7 +35,6 @@ services:
       - EMAIL_HOST_PASSWORD=
       - DEFAULT_FROM_EMAIL=noreply@siskidomain.com
       - FRONTEND_URL=http://localhost:3000
-
     depends_on:
       - db
```
mehalyna, line 39: Redis service was added but the backend service depends_on was not updated to include redis. The backend may start before Redis and fail when CACHES tries to connect.

Add - redis to the backend service depends_on list to ensure startup ordering:
```
depends_on:
  - db
  - redis
 ```

backend/api/authorization/tests/tests.py
kesshich: Done.

backend/api/authorization/views.py
kesshich: Done.

docker-compose.yml
kesshich: Done.

docker-compose.yml
kesshich: Done

mehalyna (approved):

redrembo24680 (approved):

Tonnnya (approved):

PR #185, opened 2025-10-28
Data model investment track ing and portfolio snapshot models+migrations 21

redrembo24680:


backend/apps/investors/models.py
```diff
@@ -62,13 +62,76 @@ class InvestorProfile(models.Model):
...
     class Meta:
         verbose_name = "Investor Profile"
         verbose_name_plural = "Investor Profiles"
+
+# Tracking Model
+class Tracking(models.Model):
+    id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
+    investor = models.ForeignKey(User, on_delete=models.CASCADE, related_name='tracking')
+    target_type = models.CharField(choices=[('startup', 'startup'), ('project', 'project')], max_length=32)
+    target_id = models.UUIDField()  # FK via generic relation or separate FK fields
```
mehalyna, line 86: target_id = models.UUIDField() assumes tracked targets use UUID PKs. However, StartupProfile in startups still uses BigAutoField (integer PK) per startups/migrations/0001_initial.py. This leads to value/type mismatches and runtime errors when target_type == 'startup'.

Use a content-type/generic relation or a string-typed object_id that can store both integer and UUID, e.g.:
```
from django.contrib.contenttypes.fields import GenericForeignKey
from django.contrib.contenttypes.models import ContentType

content_type = models.ForeignKey(ContentType, on_delete=models.CASCADE)
object_id = models.CharField(max_length=36)  # stores either int or uuid hex
target = GenericForeignKey('content_type', 'object_id')
```
Or, if you intend to only track projects (UUID) and not startups, restrict target_type choices and validate accordingly.

backend/apps/projects/migrations/0001_initial.py
mehalyna, line 1: The migration rewrites Project primary key from an integer BigAutoField to UUIDField. Additionally several downstream migration files were removed (multiple remove file entries). Changing PK type and rewriting migration history is high risk: it can cause data loss, broken foreign keys, and prevent upgrading existing databases.
From my point of view, it'll be better to use another approach.
Revert history rewrite. Implement PK migration in safe, multi-step migrations: 
1) add a nullable UUID field, 
2) populate UUIDs for existing rows, 
3) change FKs on dependent models to reference the UUID (add new FK fields and copy data), 
4) swap PKs with a controlled AlterField + data migration step. Keep old migrations intact and add incremental migrations instead of deleting them.

backend/apps/investors/models.py
```diff
@@ -2,6 +2,8 @@
 from django.core.exceptions import ValidationError
 from django.db import models
 from phonenumber_field.modelfields import PhoneNumberField
+from apps.projects.models import Project
```
mehalyna: This import introduces strong coupling and can produce circular import issues during Django startup and migrations. Models should generally reference other models by app label string to avoid import-time problems.

Replace direct import with lazy string reference in ForeignKey:
```
project = models.ForeignKey('projects.Project', on_delete=models.PROTECT, related_name='investments')
```
Remove the direct import to avoid circular import and keep module boundaries clean.

backend/apps/investors/models.py
redrembo24680: Yaroslav replaced id in other models with the uuid, so please check if everything is okay now.

backend/apps/investors/models.py
```diff
@@ -65,13 +66,75 @@ class InvestorProfile(TimeStampedModel):
...
+class Investment(TimeStampedModel):
+    investor = models.ForeignKey(User, on_delete=models.CASCADE, related_name='investments')
+    project = models.ForeignKey('projects.Project', on_delete=models.PROTECT, related_name='investments')
+    status = models.CharField(choices=[('committed', 'committed'), ('transferred', 'transferred'), ('returned', 'returned'), ('cancelled', 'cancelled')], default='committed', max_length=50)
+    amount_committed = models.DecimalField(max_digits=18, decimal_places=2)
+    amount_invested = models.DecimalField(max_digits=18, decimal_places=2, default=0)
+    currency = models.CharField(max_length=3, default='UAH')
+    meta = models.JSONField(null=True)  # notes, terms, round info
+
+    class Meta:
```
butuz0: `TimeStampedModel.Meta` defines default model ordering:
`ordering = ['-created_at', '-updated_at']`
If you want to keep this ordering, your `Meta` class should inherit from `TimeStampedModel.Meta`:
`class Meta(TimeStampedModel.Meta):`
If ordering is not important for this model, your implementation is perfectly fine.

backend/apps/investors/models.py
```diff
@@ -65,13 +66,75 @@ class InvestorProfile(TimeStampedModel):
...
+    projects_count = models.IntegerField()
+    total_committed = models.DecimalField(max_digits=18, decimal_places=2)
+    total_invested = models.DecimalField(max_digits=18, decimal_places=2)
+    summary = models.JSONField()  # precomputed KPIs
+
+    def __str__(self):
+        investor_name = f"{self.investor.first_name} {self.investor.last_name}".strip()
+        return f"Snapshot for {investor_name} at {self.computed_at}"
+
+    class Meta:
```
butuz0: `TimeStampedModel.Meta` defines default model ordering:
`ordering = ['-created_at', '-updated_at']`
If you want to keep this ordering, your `Meta` class should inherit from `TimeStampedModel.Meta`:
`class Meta(TimeStampedModel.Meta):`
If ordering is not important for this model, your implementation is perfectly fine.

mehalyna (approved):

butuz0 (approved):

kesshich (approved):

PR #189, opened 2025-10-29
fix ci - linter and python version

Domanov-Vitalii:
I decided to combine two tasks in this PR because they are relatively simple, so I didn't see the need to split them into separate PRs. Changes were made regarding the matrix - I removed the matrix and left only the stable 3.10 version of Python. I also indicated an explicit path to the flake configuration in the project structure to ensure there would be no discrepancies.

That's what I thought from the beginning, but now looking back I can say that it was a fun 5 hours. Changes have been made to 
image

General settings

Backend
Containerization: The work is performed in an isolated environment (container: python:3.10), which guarantees stability.
Database: For tests, a service with PostgreSQL 14 is automatically started.
Reliability: Added a health check for the database so that tests do not start until it is ready.
Network: A connection to the database is created by using the service name db as the host (POSTGRES_HOST: db).
Testing: Django tests are run with code coverage measurement, after which the report is uploaded to Codecov.
Frontend
Environment: The current version of Node.js 20.19.4 is used, which is a requirement for new versions of Vite.
Critical fix: Implemented a workaround for the npm native dependency rolldown bug. The Force clean install step forcibly removes package-lock.json and node_modules and performs npm install "from scratch". This forces npm to correctly install dependencies for Linux environments in CI.
Build: The npm run build command is executed to compile the frontend application.
Artifacts: The finished build (frontend/dist) is saved as an artifact for later use (e.g., for deployment).


.github/workflows/ci.yml
```diff
@@ -34,11 +33,11 @@ jobs:
           pip install -r backend/requirements.txt
 
       - name: "Lint (flake8)"
-        run: pip install flake8==7.0.0 && flake8 backend/
+        run: pip install flake8==7.0.0 && flake8 --config=backend/.flake8 backend/
 
       - name: 'Tests with coverage'
         run: |
-          coverage run --source='.' manage.py test
+          coverage run --source='.' manage.py test apps.authentication apps.common apps.investors apps.startups
```
butuz0: It looks like the test command explicitly runs only a few apps.
There are other apps in the project, like `users`, `user_messages`, `dashboard`, that contain tests as well, so this setup currently skips them.

backend/backend/settings.py
```diff
@@ -110,13 +111,20 @@
...
+        "NAME": ":memory:",  
     }
+
+    class DisableMigrations:
+        def __contains__(self, item):
+            return True
+        def __getitem__(self, item):
+            return None
+
+    MIGRATION_MODULES = DisableMigrations()
```
butuz0: You can set "MIGRATE": False in test database settings instead.
https://docs.djangoproject.com/en/5.2/ref/settings/#migrate

butuz0 (changes requested):

.github/workflows/ci.yml
```diff
@@ -9,38 +9,53 @@ on:
...
+        run: psql -h db -U postgres -d testdb -c "SELECT 1;"
 
-      - name: 'Tests with coverage'
+      - name: Tests with coverage
+        env:
+          POSTGRES_DB: testdb
+          POSTGRES_USER: postgres
+          POSTGRES_PASSWORD: password
+          POSTGRES_HOST: db
+          POSTGRES_PORT: 5432
```
mehalyna: The password is set as plain PGPASSWORD environment variable and used directly in a CLI command (psql -c "SELECT 1;"). This may inadvertently log secrets and is unsafe.

Use secrets and avoid injecting passwords into logs; use .pgpass or rely on the service network with proper secrets.

.github/workflows/ci.yml
```diff
@@ -9,38 +9,53 @@ on:
...
+        working-directory: backend
         run: |
-          coverage run --source='.' manage.py test
+          coverage run --source='.' manage.py test apps.authentication apps.common apps.investors apps.startups apps.projects apps.dashboard apps.user_messages apps.users
```
mehalyna: The coverage run explicitly runs manage.py test for a fixed list of apps. New apps or tests outside these apps will not run, causing silent regressions and missing coverage.

Revert to manage.py test (project-wide) or programmatically discover apps. If you must restrict, derive the list dynamically from INSTALLED_APPS or a maintained matrix.

.github/workflows/ci.yml
```diff
@@ -51,32 +66,30 @@ jobs:
...
+          cache-dependency-path: 'frontend/package-lock.json'
 
-      - name: "Install dependencies"
-        run: npm ci
-        working-directory: frontend
+      - name: Force clean install
+        run: |
+          rm -f package-lock.json
+          rm -rf node_modules
+          npm install
```
mehalyna, line 91: Removing package-lock.json and running npm install leads to non-repeatable builds and may introduce different dependency versions on CI vs local dev. This undermines reproducibility and can hide dependency-related bugs.

Use npm ci with the lockfile committed. If you must regenerate, do it intentionally in a separate step and commit the lockfile. Replace with:
```
- run: npm ci
```

Domanov-Vitalii: Thanks, I generally agree about reproducible builds. I tried npm ci with a messed up package-lock.json (Node 20.19.4, npm 10.x) — CI consistently crashes on Vite/Rolldown with the error:

Error: Cannot find native binding '@rolldown/binding-linux-x64-gnu' (link in trace to npm optional deps bug #4828)
What I've already checked and didn't help:

working-directory: frontend and valid lockfile
node 20.19.x (requirement of Vite) and npm cache clean --force
npm ci and npm install with and without cache
The root of the problem — optionalDependencies with native Rolldown bindings: a lockfile compiled on another OS sometimes does not install the required @rolldown/binding in CI (Linux). This is a known case/regression around #4828, which Vite still mentions in its own bug report.

Therefore, the current step rm package-lock.json && npm install is a temporary workaround so that the build is not blocked.

Lockfile is formally cross-platform, but with optionalDependencies with native bindings (rolldown) npm sometimes incorrectly picks up the platform package when changing OS.
npm ci strictly reproduces what is in the lockfile, and on Linux it may not install @rolldown/binding-linux-*, which gives an error (variant of bug #4828).
Therefore, simply “roll npm ci” will not help until the lockfile is generated on the target platform.

butuz0 (approved):

.github/workflows/ci.yml
Domanov-Vitalii: - name: Tests with coverage
        working-directory: backend
        run: |
          PROJECT_APPS=$(python manage.py shell --no-startup -c "from django.conf import settings; print(' '.join(app for app in settings.INSTALLED_APPS if app.startswith('apps.') or app.startswith('api.')))" | tail -n 1)
          echo "--- Running tests for the following apps ---"
          echo "$PROJECT_APPS"
          echo "--------------------------------------------"
          
          coverage run --source='.' manage.py test $PROJECT_APPS
          coverage xml
now ci test apps in INSTALLED_APPS

kesshich (approved):

mehalyna (approved):

PR #192, opened 2025-10-31
Create the desktop UI component and route RestorePassword.tsx

LambertoKalisto:


frontend/src/pages/RestorePassword.tsx
```diff
@@ -0,0 +1,142 @@
...
+
+  async function onSubmit(e: React.FormEvent) {
+    e.preventDefault();
+    setMsg(null);
+    setError(null);
+
+    if (!password || !confirm) return setError(t.required);
+    if (password.length < 8) return setError(t.tooShort);
+    if (password !== confirm) return setError(t.mismatch);
+
```
mehalyna, line 88: The component sets an error in useEffect when uid or token are missing, but onSubmit still proceeds to call the API using possibly null uid/token values (body: JSON.stringify({ uid, token, new_password: password })). This can send malformed requests or leak undefined values to the backend.

Explicitly verify uid and token inside onSubmit and prevent the request if missing. Example:
```
if (!uid || !token) {
  setError(t.invalidToken);
  return;
}
```
Place this check before setLoading(true).

frontend/src/pages/RestorePassword.tsx
```diff
@@ -0,0 +1,142 @@
...
+  const uid = search.get("uid");
+  const token = search.get("token");
+
+  useEffect(() => {
+    if (!uid || !token) {
+      setError(t.invalidToken);
+    }
+  }, [uid, token]);
+
+  async function onSubmit(e: React.FormEvent) {
```
mehalyna, line 79: After successful password change the component navigates to "/login" unconditionally. If the app uses localized routes or query params, users may lose locale context; also if navigate target should depend on origin, it's not preserved.

Either accept a redirect query param (validate it) or use a known route helper to preserve locale. E.g. `const redirect = search.get('next') || /login${locale ? ?lang=${locale} : ''}; setTimeout(() => navigate(redirect), 3000);`

frontend/src/pages/RestorePassword.tsx
mehalyna, line 1: The code sets generic "Error", "Server error", and "Network error" messages that may not be localized and may expose or hide useful information inconsistently. Also some error messages are hardcoded English, bypassing translation table.

Use translation table t for all user-visible messages and normalize server-provided messages cautiously. Replace "Error", "Server error", and "Network error" with t.* entries or add new keys.

LambertoKalisto:
Added UID/token validation inside onSubmit — prevents API calls with missing or invalid values.
Improved redirect logic — preserves locale and optional next query param after successful password change.
Localized all error messages — replaced hardcoded "Error", "Server error", "Network error" with translation keys from t.
Cleaned up error handling — added fallback unknownError, improved code readability and consistency.
Minor UX improvements — clearer success/error messages and safer async handling.

frontend/src/pages/RestorePassword.tsx
```diff
@@ -0,0 +1,156 @@
...
+  const next = search.get("next");
+  const locale = lang;
+
+  useEffect(() => {
+    if (!uid || !token) {
+      setError(t.invalidToken);
+    }
+  }, [uid, token, t]);
+
+  async function onSubmit(e: React.FormEvent) {
```
mehalyna, line 79: The client posts to /api/auth/password-reset/confirm/ using fetch without any CSRF token, authentication header, or documented server contract. If the backend expects CSRF or other anti-forgery measures, this can fail or open CSRF vectors depending on cookie auth. Additionally, uid/token are sent from query params which may be logged or exposed.

Confirm backend expects JSON POST and how it protects CSRF. If cookies are used, include credentials: 'include' and/or attach CSRF header. Prefer sending minimal token only (not uid in URL) and avoid storing tokens in logs. Example: 
```
const res = await fetch("/api/auth/password-reset/confirm/", {
method: "POST", credentials: "include", 
// if cookies required 
headers: { "Content-Type": "application/json", "X-CSRFToken": getCsrfToken(), }, 
body: JSON.stringify({ uid, token, new_password: password }), });
```

frontend/src/pages/RestorePassword.tsx
```diff
@@ -0,0 +1,156 @@
...
+  const uid = search.get("uid");
+  const token = search.get("token");
+  const next = search.get("next");
+  const locale = lang;
+
+  useEffect(() => {
+    if (!uid || !token) {
+      setError(t.invalidToken);
+    }
+  }, [uid, token, t]);
```
mehalyna, line 77: useEffect sets error if !uid || !token, but the component still renders the password form unless error exactly equals t.invalidToken (checked later). There is possible race/ordering where error may not prevent form usage; also uid/token could be present then removed. This ambiguous flow may result in attempting submit with missing uid/token checked later but flow is confusing.

Block rendering the form entirely if uid or token are missing. For example return early (like currently done) based on a derived boolean: if (!uid || !token) return invalid token UI. Move the check before hooks that depend on them.

frontend/src/pages/RestorePassword.tsx
```diff
@@ -0,0 +1,156 @@
...
+    try {
+      const res = await fetch("/api/auth/password-reset/confirm/", {
+        method: "POST",
+        headers: { "Content-Type": "application/json" },
+        body: JSON.stringify({ uid, token, new_password: password }),
+      });
+
+      if (res.ok) {
+        setMsg(t.success);
+        const redirect = next
```
mehalyna: The next query param is used directly to build redirect target after success. An attacker could craft links that redirect users to arbitrary external sites.

Validate next against an allowed internal path whitelist or ensure it begins with '/' and not '//' or include host validation. Example: 
```
const redirect = next && next.startsWith('/') ? next : /login?lang=${locale};
```

frontend/src/pages/RestorePassword.tsx
```diff
@@ -0,0 +1,156 @@
...
+          : locale
+          ? `/login?lang=${locale}`
+          : "/login";
+        setTimeout(() => navigate(redirect), 3000);
+      } else if (res.status === 400) {
+        const data = await res.json().catch(() => ({}));
+        if (data?.token || data?.uid) setError(t.invalidToken);
+        else if (data?.new_password) setError(t.weak);
+        else setError(t.unknownError);
+      } else {
```
mehalyna: The branch for res.status === 400 does await res.json().catch(() => ({})); other non-JSON errors or other 4xx codes are not parsed, possibly swallowing useful server messages. Also status codes like 401/403 not handled.

Handle other 4xx/5xx explicitly and parse body safely. Example: 
```
const text = await res.text(); 
let data = {}; 
try { data = JSON.parse(text); 
} catch {} 
// then handle data.detail || data.new_password etc. 
```
Also handle 401/403.

LambertoKalisto:
Add 401/403 errors. Delete useEffects.

mehalyna (approved):

Yenot52 (approved):

Tonnnya (approved):

PR #193, opened 2025-10-31
Why worth grid 109

Yenot52:
tests
WhyWorth


frontend/src/App.css
```diff
@@ -50,6 +50,7 @@ html, body {
   display: flex;
   flex-direction: column;
   min-height: 100vh;
+  background-color: #FFFFFF;
```
mehalyna: Adding background-color: #FFFFFF; directly to html, body enforces a white background app-wide and can break dark-mode or theming expectations and cause visual regression across many pages.

Use a CSS variable or root class for theming and avoid forcing white globally. Example: 
```
:root { 
--bg-color: #FFFFFF; 
} 
html, body { 
background-color: var(--bg-color); 
} 
```
Or restrict to .app-container instead of html/body if intended locally.

frontend/src/App.tsx
```diff
@@ -2,16 +2,19 @@ import { BrowserRouter as Router, Routes, Route } from "react-router-dom";
...
+import WhyWorthGrid from "./components/WhyWorthGrid/WhyWorthGrid";
 import './components/Header/Header.scss';
-import "./App.css"; 
+import "./App.css";
 
 function App() {
   return (
     <Router>
-      <div className="app-container">
+      <div>
```
mehalyna: Replacing ```<div className="app-container">``` with a plain `<div>` likely removes layout CSS applied to .app-container, causing layout regressions across the app.

Restore the class or move required styles accordingly. Change back to: 
```<div className="app-container">``` or update CSS to target the new structure.

File: frontend/src/App.tsx [hunk 3]
```diff
@@ -2,16 +2,19 @@ import { BrowserRouter as Router, Routes, Route } from "react-router-dom";
 import { routes } from "./routes";
 import Footer from "./components/Footer/Footer";
 import Header from "./components/Header/Header";
+import WhyWorthGrid from "./components/WhyWorthGrid/WhyWorthGrid";
```
mehalyna: Importing WhyWorthGrid directly at top-level loads its code and SCSS into the main bundle, increasing initial load; component seems non-essential content and should be lazy-loaded.

Use React.lazy + Suspense to code-split: `const WhyWorthGrid = React.lazy(() => import('./components/WhyWorthGrid/WhyWorthGrid')); ... <Suspense fallback={null}><WhyWorthGrid/></Suspense>`

frontend/src/App.css
anastasia-tebrii: background-color: #FFFFFF; removed. 
Implemented CSS variable example as suggested.

frontend/src/App.tsx
anastasia-tebrii: app-container restored

frontend/src/App.tsx
anastasia-tebrii: import { Suspense, lazy } from "react"; - added

mehalyna (approved):

kesshich (approved):

Tonnnya (approved):

PR #201, opened 2025-11-06
Profile endpoint

kesshich:
Implemented task #56 Profile Read/Update endpoint, added new directory for profiles.


backend/apps/profiles/permissions.py
```diff
@@ -0,0 +1,19 @@
...
+            return True
+        return request.user and request.user.is_authenticated
+
+    def has_object_permission(self, request, view, obj):
+        """Check if user has permission to access this specific projects"""
+        if request.method in permissions.SAFE_METHODS:
+            if not request.user.is_authenticated:
+                return False
+            return False
+        return obj.user == request.user
```
mehalyna, line 17: The object-level permission always returns False for SAFE_METHODS (the code returns False unconditionally), so even authenticated users cannot retrieve profiles via GET (or any safe methods). This breaks read access and will make API endpoints unusable for normal profile reads.
Fix logic to return True for safe methods when allowed (e.g., allow GET for authenticated users or for public profiles).
```
def has_object_permission(self, request, view, obj):
    if request.method in permissions.SAFE_METHODS:
        # allow read for authenticated users; if public read is needed implement visibility check
        return request.user and request.user.is_authenticated
    return getattr(obj, 'user', None) == request.user
```

backend/apps/authentication/views.py
```diff
@@ -10,19 +10,158 @@
...
+
+        response.set_cookie(
+            key="refresh_token",
+            value=str(tokens),
+            httponly=True,
+            secure=False,  # While still in development, True when in prod.
+            samesite="Strict",
+        )
+
+        return response
```
mehalyna, line 93: After authenticate(...) the code does not check user.is_active or any email-verified flag before generating access/refresh tokens. This allows disabled/unverified users to obtain JWTs and access protected resources.
Reject login if user is inactive or not verified:
```
if not user.is_active:
    return Response({"error": "Account inactive."}, status=status.HTTP_403_FORBIDDEN)
# If you have a is_email_verified flag:
if hasattr(user, 'is_email_verified') and not user.is_email_verified:
    return Response({"error": "Email not verified."}, status=status.HTTP_403_FORBIDDEN)
```

backend/apps/authentication/views.py
```diff
@@ -10,19 +10,158 @@
...
+            samesite="Strict",
+        )
+
+        response.set_cookie(
+            key="refresh_token",
+            value=str(tokens),
+            httponly=True,
+            secure=False,  # While still in development, True when in prod.
+            samesite="Strict",
+        )
```
mehalyna, line 91: response.set_cookie(... secure=False ...) is hard-coded and insecure for production. If this code reaches prod with secure False, tokens can be exposed over plain HTTP.
Use configuration driven flag and set Secure/SameSite/Domain/Expires intentionally:
```
secure_flag = getattr(settings, "AUTH_COOKIE_SECURE", True)
response.set_cookie(..., secure=secure_flag, samesite=getattr(settings, "AUTH_COOKIE_SAMESITE", "Lax"))
```

backend/apps/authentication/views.py
```diff
@@ -10,19 +10,158 @@
...
+        try:
+            token.blacklist()
+            response = Response(status=status.HTTP_204_NO_CONTENT)
+
+            response.delete_cookie("refresh_token")
+            response.delete_cookie("access_token")
+
+            return response
+        except Exception:
+            return Response(status=status.HTTP_400_BAD_REQUEST)
```
mehalyna, line 117: The code does token = RefreshToken(refresh_token) before the try/except block. If refresh_token is invalid or missing this constructor can raise and crash the view (uncaught), causing 500 errors.
Move construction inside try/except and handle invalid tokens:
```
refresh_token = request.COOKIES.get("refresh_token")
if not refresh_token:
    return Response(status=status.HTTP_400_BAD_REQUEST)
try:
    token = RefreshToken(refresh_token)
    token.blacklist()
except TokenError:
    return Response(status=status.HTTP_400_BAD_REQUEST)
```

backend/apps/authentication/throttling.py
```diff
@@ -1,15 +1,30 @@
+import hashlib
+
 from django.conf import settings
 from django.core.cache import cache
 from rest_framework.throttling import BaseThrottle
 
-RATE = getattr(settings, "COMMON_REDIS_THROTTLE_RATE", 6)
+RATE = getattr(settings, "COMMON_REDIS_THROTTLE_RATE", 60)
```
mehalyna: RATE = ... is evaluated at import time. Tests that use override_settings to change COMMON_REDIS_THROTTLE_RATE won't affect this module-level constant if the module is already imported — causing flaky tests and surprising runtime behavior.
Read settings inside methods or use property access so updates are respected:
```
def _get_rate(self):
    return getattr(settings, "COMMON_REDIS_THROTTLE_RATE", 60)

def allow_request(...):
    rate = self._get_rate()
    ...
```

backend/apps/authentication/tests/test_login_logout_auth.py
```diff
@@ -131,3 +131,47 @@ def test_resend_throttled(self):
         response = self.client.post(self.resend_url, resend_data, format="json")
 
         self.assertEqual(response.status_code, status.HTTP_429_TOO_MANY_REQUESTS)
+
+
+@override_settings(COMMON_REDIS_THROTTLE_RATE=6)
```
mehalyna, line 136: @override_settings(COMMON_REDIS_THROTTLE_RATE=6) assumes throttling uses settings at test runtime. Because throttling reads a module-level RATE at import-time (see previous comment), the test override likely has no effect, causing test to be flaky or always passing/failing.
Fix throttling to read settings at call-time (see previous) and/or set RATE in test by reloading module or patching the attribute directly.

backend/apps/authentication/views.py
kesshich: Done.

backend/apps/authentication/views.py
kesshich: Done.

backend/apps/authentication/views.py
kesshich: Done.

backend/apps/authentication/throttling.py
kesshich: Done.

backend/apps/authentication/tests/test_login_logout_auth.py
kesshich: _get_rate(self) method was implemented inside of throttling.py file, so know it should work fine.

backend/apps/profiles/permissions.py
kesshich: Why do we need to get user through getattr() function? And not just obj.user?

The other logic is fixed.

mehalyna: Good question - thanks for catching that!
I used `getattr()` as a safety/net because not every profile-like object in our codebase is guaranteed to expose the owner on the same attribute name (user vs owner vs something else). Using getattr(obj, 'user', None) avoids AttributeError if obj doesn't have .user and lets us fall back to another attribute or None.
If you are certain that all objects handled by this permission always have a .user property, then obj.user is fine (and slightly clearer). Otherwise I prefer a small tolerant lookup so the permission won’t crash on mixed models.

mehalyna (approved):

butuz0 (approved):

Tonnnya (approved):

PR #206, opened 2025-11-08
spam protection, rate limits, csrf/cors docs

Domanov-Vitalii:
1. Bot Protection with Google reCAPTCHA v2
Implemented Google reCAPTCHA v2 on all public-facing forms to prevent automated submissions.
Frontend: The reCAPTCHA widget is now present on the Startup Registration, Investor Registration, and Password Reset pages.
Backend: A ReCaptchaV2Field has been added to the corresponding serializers to validate the token on the server-side, ensuring that requests are legitimate.
2. Backend Rate Limiting (Throttling)
Per-Email Throttling: Limits the number of registration or password reset attempts for a single email address, even if requests come from different IPs ([EmailThrottle](vscode-file://vscode-app/e:/VSCode/resources/app/out/vs/code/electron-browser/workbench/workbench.html)).
These policies have been applied to the registration (/api/auth/register/), login, and password reset endpoints.
3. Hardened CORS Policy
The backend's Cross-Origin Resource Sharing (CORS) policy has been configured to be more secure.
CORS_ALLOW_ALL_ORIGINS is now set to False.
The system now only allows requests from the specific frontend origin defined in the FRONTEND_URL environment variable, preventing unauthorized domains from interacting with the API.
4. Environment Secrets Documentation
Template environment files have been created to document all required secrets for both services.
env.templates Documents backend secrets, including RECAPTCHA_PRIVATE_KEY, database credentials, and throttling settings.
env.templates: Documents frontend secrets, specifically VITE_RECAPTCHA_PUBLIC_KEY.
5. Security Runbook and Monitoring Documentation
A new security document has been created at registration.md.
This document provides:
Monitoring Guidelines: Instructions and sample SQL queries for detecting unusual signup spikes.
Incident Runbook: A step-by-step plan for responding to a suspected bot attack, including how to temporarily tighten rate limits and identify malicious accounts.
image
image
image
image
image


backend/apps/authentication/serializers.py
mehalyna, line 1: You added recaptcha = ReCaptchaV2Field() to multiple serializers. This makes the token mandatory and will cause requests from clients that don't include the token to fail validation (400). The PR only updates some front-end registration pages, but not password reset pages (see frontend diff), so password reset requests will break.
Either make the field optional while rolling out, or add clear client-side changes for all affected endpoints. For optional:
```
recaptcha = ReCaptchaV2Field(required=False)
```
Or validate presence conditionally (e.g., only in production) and add clear API docs and feature flags.

backend/apps/authentication/throttling.py
mehalyna, line 1: CommonRedisThrottle defines def get_cache_key(self, request): while EmailThrottle defines def get_cache_key(self, request, view):. DRF expects get_cache_key(self, request, view) — calling CommonRedisThrottle.get_cache_key(request, view) will raise a TypeError due to a missing parameter.
Standardize signatures to match DRF's expected API:
```
def get_cache_key(self, request, view):
    ...
```
Update CommonRedisThrottle to accept view (even if unused).

backend/api/authorization/throttling.py
```diff
@@ -26,3 +28,32 @@ def allow_request(self, request, view):
             current = cache.get(key) or 0
             cache.set(key, int(current) + 1, timeout=DURATION)
         return True
+
+class EmailThrottle(BaseThrottle):
+    def get_cache_key(self, request, view):
+        email = request.data.get("email")
```
mehalyna: `email = request.data.get("email")` assumes the request body contains an email field. Several throttled endpoints (login, verify by uid/token) may not include this field, leading to None and effectively disabling email throttling (method returns None -> throttle not applied) or inconsistent behavior. Also request.data access may be expensive or unavailable for some request content types.
Normalize extraction to handle multiple cases (JSON body, form-encoded, query params) and explicitly document which endpoints are covered. Example:
```
email = (
  (getattr(request, 'data', {}) or {}).get('email')
  or request.query_params.get('email')
  or request.data.get('username')  # if login uses username/email field
)
```
Consider adding tests for each endpoint to verify throttle behavior.

frontend/src/pages/PasswordResetRequest.tsx
mehalyna, line 1: You added recaptcha to server-side password reset serializers but did not add ReCAPTCHA widget or token submission in this frontend page. Password reset submissions will fail validation.
Add ReCAPTCHA component and include token in POST body, similar to other updated pages, or remove server-side requirement until all clients are updated. Example:
```
<ReCAPTCHA sitekey={import.meta.env.VITE_RECAPTCHA_PUBLIC_KEY} onChange={token => setRecaptchaToken(token)} />
...
formData.append('recaptcha', recaptchaToken)
```

backend/api/authorization/throttling.py
Domanov-Vitalii: now, EmailThrottle cover 
/api/auth/register/
/api/auth/password-reset/
/api/auth/login/
/api/auth/verify-email/uid/token/
/api/auth/password-reset-confirm/uid/token/
because of new mechanism extractions uid

frontend/src/pages/PasswordResetRequest.tsx
Domanov-Vitalii: The functionality should be working as expected. Now implementation for the password reset request page already matches the backend requirement.

backend/apps/authentication/throttling.py
Domanov-Vitalii: implemented

backend/apps/authentication/serializers.py
Domanov-Vitalii: You are absolutely right. I`d updated the PasswordResetRequest.tsx component to include the reCAPTCHA widget and send the token with the request.

mehalyna (approved):

redrembo24680 (approved):

butuz0 (approved):

PR #207, opened 2025-11-08
#95

### LINKED ISSUE #95: Frontend: Password reset UI (request + new password)
**Title:** frontend: implement Password Reset flow (desktop UI for request & confirm)

**Description:**
Implement two screens: request reset (enter email) and set new password (token in URL). Provide UX messages per mockup and validation for new password.

**Checklist**

* [x] `PasswordResetRequest` component: form to POST `/api/auth/password-reset/` (mockable).
* [x] `PasswordResetConfirm` component: route `/reset-password/?token=...`, form to POST `/api/auth/password-reset/confirm/`.
* [x] Validate password strength & confirm password field.
* [x] Show success page instructing user to check email (always show success after POST).
* [x] Tests for flows and validation.

**Acceptance criteria**

* Request endpoint always returns success UI (no account enumeration).
* Confirm with valid token triggers success UI and allows login.
* Invalid/expired token shows helpful error and link to resend.
* Tests pass.

---
**Original Issue:** https://github.com/Project-Stage-Academy/Forum-PrSt-FullStack/issues/27
**Created by:** mehalyna
**Original creation date:** 2025-10-01T12:39:10Z

Yura8q:
I added PasswordResetConfirm and test. Also fixed and added everything that was asked in the task. Easily fixed App.tsx and added so that it was visible visually


frontend/src/App.tsx
mehalyna, line 1: The PR adds route entries in one `<Routes>` instance and then introduces a second `<Routes>` block later in the file. Having multiple overlapping Routes and duplicate route declarations (including a wildcard 404) can cause unexpected routing behavior, surprising navigation, and makes the app harder to reason about.
Consolidate all route declarations into a single `<Routes>` component (or import/consume the centralized routes array). Remove the duplicated `<Routes>` block and ensure the new routes are added to the canonical router location. Example merge:
```
<Routes>
  <Route path="/forgot-password" element={<PasswordResetRequest />} />
  <Route path="/reset-password" element={<PasswordResetConfirm />} />
  {/* other routes */}
  <Route path="*" element={<div>404 Not Found</div>} />
</Routes>
```

frontend/src/pages/PasswordResetRequest.tsx
```diff
@@ -39,42 +39,46 @@ export default function PasswordResetRequest({ lang = "uk" }: Props) {
...
-            method: "POST",
-            credentials: "include",
-            headers: {
-                "Content-Type": "application/json",
-                ...(csrftoken ? { "X-CSRFToken": csrftoken } : {}),
-            },
-            body: JSON.stringify({ email: v }),
-  });
+      const getCookie = (name: string) => {
+        const row = document.cookie.split("; ").find((r) => r.startsWith(name + "="));
```
mehalyna: The code now uses document.cookie.split(";").find(r => r.startsWith(name + "=")). When cookies are separated by "; " the resulting tokens often begin with a leading space (e.g. " csrftoken=...") and startsWith will fail, causing the CSRF cookie to be missed intermittently.
Normalize trimming and robustly parse cookies. Example:
```
const getCookie = (name: string) =>
  document.cookie.split(";").map(c => c.trim()).find(c => c.startsWith(name + "="))
    ?.split("=").slice(1).join("=") || undefined;
```

frontend/src/pages/PasswordResetRequest.tsx
```diff
@@ -39,42 +39,46 @@ export default function PasswordResetRequest({ lang = "uk" }: Props) {
...
+        return row ? decodeURIComponent(row.split("=")[1]) : undefined;
```
mehalyna: Using row.split("=")[1] will drop portions of the value when the cookie value itself contains "=" (common for base64 or signed cookies). This leads to truncated CSRF token values and failed requests.
Use .split("=").slice(1).join("=") to preserve "=" in values. Example:
```
return row ? decodeURIComponent(row.split("=").slice(1).join("=")) : undefined;
```

frontend/src/pages/PasswordResetRequest.tsx
```diff
@@ -39,42 +39,46 @@ export default function PasswordResetRequest({ lang = "uk" }: Props) {
...
+      };
+
+      const meta = document.querySelector('meta[name="csrf-token"]') as HTMLMetaElement | null;
+      const csrftoken = getCookie("csrftoken") ?? meta?.content;
```
mehalyna: The code retrieves a csrftoken from cookies/meta but the diff does not show it being included in the outgoing fetch headers. If required by the backend, omission will cause CSRF validation failures. Also relying only on meta or cookie without sending the header is insufficient.
Ensure CSRF token is attached to requests, e.g. add:
```
headers: {
  "Content-Type": "application/json",
  ...(csrftoken ? { "X-CSRFToken": csrftoken } : {}),
}
```

mehalyna (approved):

redrembo24680 (approved):

LambertoKalisto (approved):

PR #209, opened 2025-11-09
add endpoint for landing page

redrembo24680:


backend/backend/urls.py
mehalyna, line 1: apps.common.urls is included under both common/ and api/content/. This exposes the same views under two different prefixes, causing duplicate endpoints, possible client confusion, and harder-to-maintain routing.
Choose a single canonical include or namespace them clearly. For example, prefer API prefix and remove the other include, or namespace:

backend/apps/common/tests/test_landing_page.py
```diff
@@ -0,0 +1,41 @@
+from rest_framework import status
+from rest_framework.test import APITestCase
+
+
+class LandingContentAPITests(APITestCase):
+
+    def setUp(self) -> None:
+        self.url = "/api/content/landing/"
```
mehalyna: The test hardcodes self.url = "/api/content/landing/". That will break if routing changes or if the canonical prefix is changed (see duplicate includes). Tests should rely on URL names to remain stable across refactors.
Use Django's reverse() with the named URL. Ensure apps.common.urls is namespaced if needed:
```
from django.urls import reverse
self.url = reverse("landing-content")  # or reverse("common:landing-content") if namespaced
```

backend/backend/urls.py
redrembo24680: I have added two separate urlpatterns for common and content

kesshich (approved):

Domanov-Vitalii (approved):

VovaD199 (approved):

PR #210, opened 2025-11-09
Frontend: Reset Confirm Page (desktop) #86

LambertoKalisto:
Add PasswordResetConfirm Page with reads uid and token from URL query parameters


frontend/src/pages/PasswordResetConfirm.tsx
```diff
@@ -0,0 +1,109 @@
...
+          />
+
+          {msg && <div style={{ color: "#0f7b0f", marginBottom: 8 }}>{msg}</div>}
+
+          <button type="submit">{t.submit}</button>
+        </form>
+      )}
+    </div>
+  );
+}
```
mehalyna, line 109: The POST body sends { uid, token, new_password }. Many backends implementing Django/DRF password reset expect uidb64 and token or different field names (or expect password vs new_password). If the property names differ the reset will always fail; this becomes a breaking runtime bug.
Confirm backend contract and align keys. Example if backend expects uidb64 and new_password:
```
body: JSON.stringify({ uidb64: uid, token, new_password: password })
```
Or if it expects password instead of new_password:
```
body: JSON.stringify({ uid, token, password })
```
Add unit/integration tests that mock the API to catch mismatch.

frontend/src/routes.tsx
```diff
@@ -17,6 +18,7 @@ export const routes = [
   { path: "/startups/:id", element: <StartupView /> },
   { path: "/dashboard", element: <InvestorDashboard /> },
   { path: "/messages", element: <Inbox /> },
-  { path: "/forgot-password", element: <PasswordResetRequest /> },
+  { path: "/reset-password", element: <PasswordResetRequest /> },
```
mehalyna: Changing the public route path from /forgot-password to /reset-password is an externally visible API change. If backend-generated reset emails or other places still link to /forgot-password, users will receive broken links.
Verify and update all places that generate or reference the reset URL (backend email templates, environment config, documentation). Consider adding a redirect from the old path to the new one to maintain backward compatibility

mehalyna (commented):
Routes are defined inline in App.tsx and also exported as an array in routes.tsx. The PR touches both, increasing risk they diverge (already happened: forgot-password vs reset-password naming). This creates maintenance and consistency problems (single responsibility violation for routing).
Consolidate routing to a single source of truth. Either import and render routes from routes.tsx in App.tsx, or keep inline routes but remove/unused routes.tsx. 
```
import { routes } from "./routes";
...
<Routes>
  {routes.map(r => <Route key={r.path} path={r.path} element={r.element} />)}
</Routes>
```
Add tests that assert route coverage.

LambertoKalisto:
Change routes in App.tsx
Change name in routes.tsx
Change uid name in api responce

Yenot52 (approved):

mehalyna (approved):

Yura8q (approved):

PR #219, opened 2025-11-10
imlemented track/untrack endpoints for startups and projects and tests

Domanov-Vitalii:
https://github.com/Project-Stage-Academy/UA-4252-Python-FS/issues/20 - linked issue

Endpoints
POST /api/tracking/: implemented via apps/investors/views.py - TrackingViewSet. Idempotency handled (IntegrityError - return existing with 200). Serializer: TrackingCreateSerializer validates target existence and disables unique-together validator for idempotent flow.
DELETE /api/tracking/{id}/: implemented via DestroyModelMixin in TrackingViewSet with owner guard IsTrackingOwner. Returns 204.
GET /api/investors/{id}/tracking/: implemented as InvestorTrackingListView with pagination StartupPagination and injection of basic target data using TrackedStartupSerializer / TrackedProjectSerializer.
Permissions and validation
Auth required everywhere: get_permissions on TrackingViewSet; IsAuthenticated on list view.
Owner-only delete: IsTrackingOwner.
target_id existence validated in TrackingCreateSerializer.validate.
Tests
Covered: create, idempotent re-POST (200), invalid target (400), delete success (204), delete not owner (403), list with filtering, list not owner (403). See apps/investors/tests.py - TrackingAPITestCase.


backend/apps/investors/permissions.py
```diff
@@ -0,0 +1,9 @@
+from rest_framework import permissions
+
+class IsTrackingOwner(permissions.BasePermission):
+    """
+    Дозвіл, який перевіряє, чи є поточний користувач
+    власником 'Tracking' об'єкта.
+    """
```
VovaD199, line 7: Please translate the docstring into Ukrainian.

backend/apps/investors/serializers.py
```diff
@@ -0,0 +1,106 @@
...
+
+        return data
+
+    def create(self, validated_data):
+        """
+        Перевизначаємо create для ідемпотентності.
+        Якщо запис вже існує (завдяки unique_together), ми його повертаємо.
+        """
+        return super().create(validated_data)
+
```
VovaD199, line 63: Add try/except IntegrityError to create() — now the comment says “idempotence”, but there is no actual handling of duplicate records.

    def create(self, validated_data):
        try:
            return super().create(validated_data)
        except IntegrityError:
            return Tracking.objects.get(**validated_data)

backend/apps/investors/urls.py
VovaD199, line 1: Add include(router.urls) after path('api/investors/', include(...)) in main urls.py
So that all routes are properly connected in /api/

VovaD199: Add `IsAuthenticated` + `IsTrackingOwner` check to `InvestorTrackingListView`
So that investors do not see other people's trackings (corresponds to `test_list_tracking_not_owner_forbidden`)

backend/backend/urls.py
VovaD199, line 1: `common/` is moved outside of `api/`, while all other endpoints are under /api/.

 Tests, including `test_landing_page.py`, expect `reverse("landing-content") `→ /api/common/landing-content/

VovaD199, line 1: projects and investors are connected simply as "api/", so routes from their urls.py are added without a prefix:
e.g. /api/tracking/, /api/investors/<id>/tracking/, /api/startups/<id>/projects/.

    path("api/projects/", include("apps.projects.urls")),
    path("api/investors/", include("apps.investors.urls")),

VovaD199 (changes requested):

backend/apps/investors/permissions.py
Domanov-Vitalii: It looks like a mistake and should have been "translate into English". So I did that.

backend/apps/investors/serializers.py
Domanov-Vitalii: Idempotence is implemented in the TrackingViewSet.create method: the request is wrapped in a transaction, in case of IntegrityError we get the existing record and return 200 OK with it, which is what the task requires.

backend/apps/investors/urls.py
Domanov-Vitalii: When i include apps/invetors/urls.py in main urls.py using path("api/", include("apps.investors.urls")), Django automatically creates the following routes:
api/tracking/ (for POST)
api/tracking/{id}/ (for DELETE)
api/investors/<uuid:investor_id>/tracking/ (for GET)
So no changes to urls.py are needed, as it is already set up correctly. The current structure is correct.

I also added ownership checking. I created a separate class IsInvestorSelf for this, since IsTrackingOwner works at the object level and is not suitable for ListView. The new class checks the investor_id from the URL, which completely solves the problem.

backend/backend/urls.py
Domanov-Vitalii: After removing the line path("common/", include("apps.common.urls")), all URLs for the common application will be accessible exclusively through the /api/ prefix, which is consistent with the project architecture and test expectations.

Domanov-Vitalii: now, all urls are configured correctly

butuz0 (approved):

VovaD199 (approved):

redrembo24680 (approved):

PR #222, opened 2025-11-12
Permissions roles publish #52 task.

kesshich:
Task #52 implemented with next profile publishing rules:
- Unauthorized users cannot publish (403).
- Publish endpoint enforces required fields and returns helpful list of missing items.
- On success, is_published true and published_at set.

Also some code was refactored, global variables are set, new migrations added due to new models fields.


backend/apps/investors/models.py
```diff
@@ -42,13 +42,18 @@
...
-    description = models.TextField()
-    investment_range_min = models.DecimalField(max_digits=12, decimal_places=2)
-    investment_range_max = models.DecimalField(max_digits=12, decimal_places=2)
+    description = models.TextField(blank=True, default="")
+    investment_range_min = models.DecimalField(
+        max_digits=12, decimal_places=2, null=True, blank=True
+    )
+    investment_range_max = models.DecimalField(
+        max_digits=12, decimal_places=2, null=True, blank=True
+    )
```
mehalyna, line 56: `investment_range_min` and `investment_range_max` have been changed to null=True, blank=True. The `clean()` method performs a direct comparison and will raise TypeError when one or both are None. Guard the comparison for None:
```
def clean(self):
    if self.investment_range_min is not None and self.investment_range_max is not None:
        if self.investment_range_max < self.investment_range_min:
            raise ValidationError("investment_range_max must be >= investment_range_min")
```

backend/backend/settings.py
```diff
@@ -26,8 +26,10 @@
 
 # SECURITY WARNING: keep the secret key used in production secret!
 SECRET_KEY = os.environ.get("DJANGO_SECRET_KEY", "unsafe-default-key")
-RECAPTCHA_PUBLIC_KEY = os.environ.get("RECAPTCHA_PUBLIC_KEY")
-DRF_RECAPTCHA_SECRET_KEY = os.environ.get("RECAPTCHA_PRIVATE_KEY")
+RECAPTCHA_PUBLIC_KEY = os.environ.get("RECAPTCHA_PUBLIC_KEY", "unsafe-recaptcha-key")
+DRF_RECAPTCHA_SECRET_KEY = os.environ.get(
+    "RECAPTCHA_PRIVATE_KEY", "unsafe-recaptcha-key"
+)
```
mehalyna: Defaulting RECAPTCHA_PUBLIC_KEY and DRF_RECAPTCHA_SECRET_KEY to fixed "unsafe-" values in settings may leak insecure defaults into deployed environments and mask misconfiguration. This can allow bypass or unexpected behavior in production.
Do not provide secret defaults; require explicit environment values or fail fast:
```
RECAPTCHA_PUBLIC_KEY = os.environ["RECAPTCHA_PUBLIC_KEY"]
DRF_RECAPTCHA_SECRET_KEY = os.environ["RECAPTCHA_PRIVATE_KEY"]
```
Or at minimum log/raise an error on missing values in non-DEBUG mode.

VovaD199 (approved):

backend/apps/investors/models.py
kesshich: Done.

backend/backend/settings.py
kesshich: Done.

butuz0 (approved):

mehalyna (approved):

redrembo24680 (approved):

PR #223, opened 2025-11-12
Frontend: About/Products/Story content components#61

### LINKED ISSUE #61: Frontend: About / Products / Story content components
**Title:** frontend: implement `About`, `ProductsList`, and `Story` components for main content

**Purpose:** Render sanitized `about_html`, structured Products/Services list, and Story block with headings and internal anchors.

**Checklist**

* [ ] `About` component: render `about_html` safely (use DOMPurify or server-side sanitization).
* [ ] `ProductsList` component: render list/cards of products/services (data either from profile response or separate `projects` endpoint).
* [ ] `Story` component: render timeline/text sections as per mockup.
* [ ] Add smooth scroll / in-page anchor support for right-side quick links (optional).
* [ ] Unit tests to ensure content rendering and sanitization.

**Acceptance criteria**

* Rich content displays correctly (paragraphs, lists, headings).
* No raw unsafe HTML executes (sanitization).
* Tests pass.

---
**Original Issue:** https://github.com/Project-Stage-Academy/Forum-PrSt-FullStack/issues/61
**Created by:** mehalyna
**Original creation date:** 2025-10-01T13:54:49Z

Yura8q:


frontend/src/components/Main/Story.tsx
```diff
@@ -0,0 +1,78 @@
...
+const Story: React.FC<Props> = ({ sections, sidebar, title = "Про компанію", showTOC = false }) => {
+  const sanitize = (html: string) => {
+    try {
+      const doc = new DOMParser().parseFromString(html ?? "", "text/html");
+      doc.querySelectorAll("script, style").forEach((el) => el.remove());
+      return doc.body.innerHTML;
+    } catch {
+      return html;
+    }
+  };
```
mehalyna: The sanitize function only strips script and style tags via DOMParser but leaves other XSS vectors (inline event handlers, javascript: URLs, SVG payloads, style attributes, etc.). Using dangerouslySetInnerHTML with this returned HTML can allow script execution or injection.
Use a well-maintained sanitizer like DOMPurify (or server-side sanitize-html) instead of manual stripping.
```
// install dompurify and @types/dompurify
import DOMPurify from "dompurify";

const sanitize = (html: string) => DOMPurify.sanitize(html || "", {ALLOWED_TAGS: /* choose policy */});
```
Replace current sanitize impl with DOMPurify and configure a strict policy (disallow inline handlers, javascript: hrefs, etc.).

frontend/src/components/Main/Story.tsx
```diff
@@ -0,0 +1,78 @@
...
+  sections: StorySection[];
+  sidebar: SidebarItem[];
+  title?: string;
+  showTOC?: boolean;
+};
+
+const Story: React.FC<Props> = ({ sections, sidebar, title = "Про компанію", showTOC = false }) => {
+  const sanitize = (html: string) => {
+    try {
+      const doc = new DOMParser().parseFromString(html ?? "", "text/html");
```
mehalyna: new DOMParser() is a browser API and may be undefined in Node/SSR environments (or some test runners), causing runtime ReferenceError or crash during server-side rendering.
Guard usage or fallback to server-safe sanitizer. For example:
```
const sanitize = (html: string) => {
  if (typeof DOMParser === "undefined") {
    // fallback to server-safe sanitizer or a no-op that escapes HTML
    return ""; // or use server sanitizer
  }
  // existing DOMParser flow...
};
```
Better: use DOMPurify with isomorphic support (jsdom DOM + DOMPurify) so it works in tests/SSR.

frontend/src/components/Main/Story.tsx
```diff
@@ -0,0 +1,78 @@
...
+      doc.querySelectorAll("script, style").forEach((el) => el.remove());
+      return doc.body.innerHTML;
+    } catch {
+      return html;
+    }
```
mehalyna: The catch block returns the original html unmodified if parsing fails, potentially reintroducing unsafe markup that the sanitizer intended to remove.
On parse failure return a safe fallback (empty string or escaped text) rather than raw input:
```
catch {
  return ""; // or escapeHtml(html)
}
```

frontend/src/components/Main/Story.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+  return (
+    <section className="story" id="story">
+      <div className="story__grid">
+        <div className="story__main">
+          {shouldShowTitle && <h2 className="story__pageTitle">{title}</h2>}
+          {sections.map((s) => (
+            <article key={s.id} id={s.id} className={`story__section ${s.highlight ? "is-highlight" : ""}`}>
+              <h3 className="story__h3">{s.heading}</h3>
+              <div className="story__content" dangerouslySetInnerHTML={{ __html: sanitize(s.html) }} />
```
mehalyna: dangerouslySetInnerHTML={{ __html: sanitize(s.html) }} is used; given the weak sanitizer and fallback behavior, this is a direct XSS vector.
Only use dangerouslySetInnerHTML with HTML that has been processed by a robust sanitizer (DOMPurify). Alternatively, render safe markdown-to-react or parse HTML into allowed React elements to avoid direct HTML insertion.

mehalyna (approved):

Yenot52 (approved):

Tonnnya (approved):

PR #230, opened 2025-11-16
add endpoint investment

redrembo24680:


Domanov-Vitalii (approved):

Yura8q (approved):

backend/apps/investors/permissions.py
```diff
@@ -14,4 +15,24 @@ class IsInvestorSelf(permissions.BasePermission):
     matches the authenticated user's ID.
     """
     def has_permission(self, request, view):
-        return request.user.id == view.kwargs.get('investor_id')
\ No newline at end of file
+        return request.user.id == view.kwargs.get('investor_id')
```
mehalyna: IsInvestorSelf checks view.kwargs.get('investor_id'), while InvestorInvestmentsListView uses id as the kwarg in the route. This inconsistency can result in permissions incorrectly denying access (or KeyErrors if code expects a value).
Make the URL kwarg name consistent (prefer investor_id) and update all references/tests to match; or read both keys defensively:
```
investor_id = view.kwargs.get('investor_id') or view.kwargs.get('id')
return request.user.id == investor_id
```

backend/apps/investors/views.py
```diff
@@ -153,4 +158,39 @@ def fetch_and_inject_targets(self, request, tracking_data):
...
             item['target'] = targets_map.get(str(item['target_id']))
 
-        return tracking_data
\ No newline at end of file
+        return tracking_data
+    
+
+class InvestmentViewSet(mixins.CreateModelMixin,
+                        mixins.UpdateModelMixin,
+                        viewsets.GenericViewSet):
```
mehalyna, line 168: The viewset mixes only CreateModelMixin and UpdateModelMixin - there is no ListModelMixin or RetrieveModelMixin. Tests and other code reverse 'investment-list' and expect list/detail routes; router registration may or may not expose these names (create exists, detail exists due to update, but listing behavior is ambiguous).
Add the missing mixins explicitly if listing or retrieving is desirable, or document that only create/patch are supported.

backend/apps/investors/serializers.py
```diff
@@ -104,4 +106,82 @@ class Meta:
...
+            raise serializers.ValidationError("Project not found.")
+
+        if project.visibility not in ['public', 'investor_allowed']: 
+            raise serializers.ValidationError("You cannot invest in this project (not public).")
+
+        FUNDRAISING_STATUS = getattr(settings, 'FUNDRAISING_STATUS', 'fundraising')
+        if project.status != FUNDRAISING_STATUS:
+            raise serializers.ValidationError(f"The project is not currently fundraising (status: {project.status}).")
+        
+        return project
```
mehalyna: The field is named project_id (UUIDField) but the validator returns a Project instance that will be placed in validated_data['project_id'] and later popped and used as project. This mixing of types (id -> instance) is surprising and may confuse future maintainers.
Rename the serializer field to project (PrimaryKeyRelatedField) or explicitly convert/assign to a project key. 
```
project_id = serializers.UUIDField(write_only=True)
def validate_project_id(self, value):
    ...
    return str(project.pk)  # or return value but then in create lookup project again
def create(self, validated_data):
    project_id = validated_data.pop('project_id')
    project = Project.objects.get(pk=project_id)
    ...
```
Or use project = serializers.PrimaryKeyRelatedField(queryset=Project.objects.all(), write_only=True).

backend/apps/investors/views.py
redrembo24680: class InvestmentViewSet(mixins.CreateModelMixin,
                        mixins.UpdateModelMixin,
                        mixins.ListModelMixin,
                        mixins.RetrieveModelMixin,
                        viewsets.GenericViewSet)
Added other mixins for all operations

mehalyna (approved):

